Issue #146, opened 2014-09-08
Try to pre-cache line starts/ends/indents prior to parse


puzrin:
It seems attractive to prefill arrays with begin/end/indent offsets for each line. That can simplify many things.

Need to benchmark it.

puzrin:
Closing - seems not needed now.

PR #158, opened 2015-01-13
Issue #146 Improvement

nomadslim:
Created a function getOffsetsArray which returns a 3D array with the
lines start, indent, end.


puzrin:
Sorry, incomplete changes are not accepted. That issue needs proofs with benchmark, that it improves parse speed.

Issue #104, opened 2013-11-17
[dumper] Unnecessary quoting for strings with internal dashes


puzrin:
Dumper add quotes to all strings that have `-`. Seems, `css` property value quotes can be skipped without problems (al least, it's ok when i delete those manually):

``` yaml
glyphs: 
  - css: "day-cloudy-gusts"
    code: 0xE800
    uid: c77de216fba51e29c6fc03012d4d5973
    search: [day, cloudy, gusts, wind, windy]
```

Can that be improved without serious yaml rewrite?

PR #181, opened 2015-04-20
Adds literal, folded, and single-quote scalar dump

### LINKED ISSUE #66: [writer] Implement full-weight scalar style resolver.
It should resolve scalars to all available YAML scalar styles: [plain](http://www.yaml.org/spec/1.2/spec.html#id2788859), [single-quoted](http://www.yaml.org/spec/1.2/spec.html#id2788097), [double-quoted](http://www.yaml.org/spec/1.2/spec.html#id2787109), [folded](http://www.yaml.org/spec/1.2/spec.html#id2796251), and [literal](http://www.yaml.org/spec/1.2/spec.html#id2795688).

Examples:
``` yaml
text: Simple string.
```
``` yaml
text: 'String with trailing spaces...  '
```
``` yaml
text: "String with unprintable (\0, \uFFFE) characters."
```
``` yaml
text: "String with \"quotation\" 'mark' characters."
```
``` yaml
# Folded style: there is only one line break after the title.
text: >
  Multiline string.

  This is very a long
  text. It contains no
  line breaks. So it
  is better to use
  folded style here.
```
``` yaml
# Literal style: there is a line break after _each_ actual line.
text: |
  This text contains
  a lot of short lines.
  So, it is better to
  use literal style here.
```

Note. The actual resolver makes only rough choose between plain and double-quoted styles.

isaacs:
Expands the ways that string values can be dumped to yaml.

Use simple scalar when possible, as before.

If simple is not possible, but single-quoted is allowed (ie, no
unprintables or escaped characters), then use that.

If there are linebreaks, and no lines have trailing or leading spaces,
then use folded or literal.  Prefer literal if all lines are below a
reasonable limit.

If neither literal nor folded are possible, then fall back to
double-quoted.

Fixes #66

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -132,6 +132,32 @@ function State(options) {
 }
 
 
+function indentString(string, spaces) {
+  var ind = common.repeat(' ', spaces),
+      position = 0,
+      next = -1,
+      result = '',
+      line,
+      length = string.length;
+
+  while (position < length) {
+    next = string.indexOf('\n', position);
+    if (next === -1) {
+      line = string.slice(position);
+      position = length;
+    } else {
+      line = string.slice(position, next + 1);
+      position = next + 1;
+    }
+    if (line.length && line !== '\n') {
+      result += ind;
+    }
+    result += line;
+  }
+
+  return result;
+}
+
 function generateNextLine(state, level) {
   return '\n' + common.repeat(' ', state.indent * level);
 }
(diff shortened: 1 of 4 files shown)


isaacs:
review, please: @nodeca

The `StringBuilder` class is probably unnecessary, since it turned out that it was simpler to just use booleans for the other dump types.  But I left it in since it's a bit nicer way to organize that logic.  I'm happy to refactor it out if you feel it's not worth it.

puzrin:
How about speed change of new dumper? I don't ask about every case, only about situation in general.

/cc @dervus please take a look

isaacs:
@puzrin Do you have a benchmark script that it could be tested against?

I try to avoid making optimizations without having some way of measuring specific impact.  It's your project, so I'm happy to polish it as much as you'd like.  But most of the "slow path" things will only start to be slow once you have a very long multi-line string, which we already have buffered in memory, and the use of slice and string appending takes advantage of the cons string capabilities in V8.  Comparing characters vs charcodes is not really a significant difference; it's only 0.39% faster in my micro-benchmarks, on an operation that can be performed at about 10,000 ops/ms on my machine.  A slightly greater gain could be done by simply moving the character code vars inline in the functions that use them, rather than defining them at the module-level.  (Also, those should be `const`s, which will get us some more speed improvements, at the cost of losing older runtimes.  Maybe a TODO for later?)

Much greater gains could be made by tracking down deopts and inline opportunities.

isaacs:
Sorry, I was measuring wrong.  Not 10,000 ops/ms.  200,000 ops/ms.

Character and integer comparisons are *crazy fast* in v8!

puzrin:
@isaacs there are no separate dumper benchmark (nobody cared much). We have only https://github.com/nodeca/js-yaml/tree/master/benchmark and extend it when needed.

puzrin:
@isaacs i've finished my review pass (generic design/logic). LGTM.  It worth to wait @dervus, because he is more experienced in yaml specs. He told me he will look carefully tommorow.

Please, squash into single commit.

dervus:
The first thing I noticed is the test. Most likely you want to create individual test for that, since test/samples-common do not check formatting; just values of load and dump-load. Will take a look at the rest tomorrow. Thanks for such massive PR anyway!

dervus:
IMO it should be `test/units/dump-scalar-styles.js`

isaacs:
@dervus @puzrin Thank you both for the review.  I'm looking forward to any new stuff you find :)

I've updated a few more things, and added a unit test.

lib/js-yaml/dumper.js
```diff
@@ -132,6 +132,30 @@ function State(options) {
 }
 
 
+function indentString(string, spaces) {
+  var ind = common.repeat(' ', spaces);
+  var position = 0;
+  var next = -1;
+  var result = '';
+  var line;
+  var length = string.length;
```
dervus: Please use single `var` statement for the whole block of assignments. See any other place in our code for a reference.

lib/js-yaml/dumper.js
```diff
@@ -132,6 +132,30 @@ function State(options) {
...
+      position = length;
+    } else {
+      line = string.slice(position, next + 1);
+      position = next + 1;
+    }
+    if (line.length && line !== '\n') {
+      result += ind;
+    }
+    result += line;
+  }
```
dervus: Please always wrap code blocks with empty lines.

lib/js-yaml/dumper.js
```diff
@@ -132,6 +132,30 @@ function State(options) {
...
+  return result;
+}
+
```
dervus: Need 2 empty lines between top-level definitions.

lib/js-yaml/dumper.js
```diff
@@ -150,81 +174,331 @@ function testImplicitResolving(state, str) {
...
+  var indent = state.indent * level;
+  var max = 80;
+  if (indent < 40) {
+    max -= indent;
+  } else {
+    max = 40;
+  }
+
+  for (var position = 0; position < object.length; position++) {
```
dervus: All `var` statements must be declared at top-level of function body. Also it's preferred, but not required to declare them at very beginning of the function body.

isaacs:
@dervus Thanks, I'll update those style points soon.

Aside on that, would it be worthwhile to use a linter that catches these kinds of things as well?  I think jshint and eslint are both pretty popular, and have more options.  Jslint is a bit behind the times feature-wise, I think.

lib/js-yaml/dumper.js
```diff
@@ -150,81 +174,331 @@ function testImplicitResolving(state, str) {
...
+  } else {
+    folded = true;
+    literal = true;
+  }
+
+  if (maybeNum) {
+    simple = false;
+  }
+
+  var single = true;
```
dervus: This is called "plain" in YAML, but if you think "simple" is better, I do not insist.

puzrin:
@isaacs i've already migrated most of projects to eslint. This happens when repo activity rises. Will update js-yaml too after your PR megre.

lib/js-yaml/dumper.js
```diff
@@ -150,81 +174,331 @@ function testImplicitResolving(state, str) {
...
-          CHAR_PERCENT              === character ||
-          CHAR_COMMERCIAL_AT        === character ||
-          CHAR_COLON                === character ||
-          CHAR_GRAVE_ACCENT         === character) {
-        isQuoted = true;
+    state.dump = "'" + object + "'";
+    return;
+  }
+
+  var simple = true;
```
dervus: This is called "plain" in YAML, but if you think "simple" is better, I do not insist.

lib/js-yaml/dumper.js
```diff
@@ -150,81 +174,331 @@ function testImplicitResolving(state, str) {
...
+  var first = object.length ? object.charCodeAt(0) : 0;
+  var spaceWrap = (CHAR_SPACE === first ||
+                   CHAR_SPACE === object.charCodeAt(object.length - 1));
+
+  var maybeNum = (CHAR_MINUS    === first ||
+                  CHAR_QUESTION === first);
```
dervus: I don't quite understand the purpose of this.

lib/js-yaml/dumper.js
```diff
@@ -150,81 +174,331 @@ function testImplicitResolving(state, str) {
   return false;
 }
 
-function writeScalar(state, object) {
-  var isQuoted, checkpoint, position, length, character, first;
+function StringBuilder (source) {
```
dervus: Do not put a whitespace after function names.

dervus:
Other than style issues it seems okay for me.

lib/js-yaml/dumper.js
isaacs: It's a carry-over from the previous dumper implementation that would not dump a string as a simple (unquoted) scalar if it started with `-` or `?`.  I'm not sure why.  I think it is a numeric format thing?

isaacs:
@dervus That `CHAR_SPACE || CHAR_QUESTION` thing was added in https://github.com/nodeca/js-yaml/commit/b946fcdc6edee8da19ab7bc5c5fa01e9ba03eb40 by @puzrin to close #104.  I'm not sure what cases it was intended to retain quoting for, but I didn't want to change behavior for that situation.

Here's some edge cases that might be relevant:

```
> require('./').safeLoad('---asdf')
'asdf'
> require('./').safeLoad('--asdf')
'--asdf'
> require('./').safeLoad('--- asdf')
'asdf'
> require('./').safeLoad('??? asdf')
'??? asdf'
> require('./').safeLoad('??? --- asdf')
'??? --- asdf'
> require('./').safeLoad('? --- asdf')
{ '--- asdf': null }
```

It seems like quoting things that start with `?` or `-` is prudent, since it might be some magic yaml stuff.  Perhaps I was misunderstanding the intent, though, and the name "maybeNum" is actually the only problem.  I'll rename it to `maybeYamlMagic` or something :)

lib/js-yaml/dumper.js
dervus: Ah! Yeah, but it's actually not about numbers. Number-looking strings are secured by the implicit resolving check. It's actually a simplified check for "? " and "- " character sequences that denote mappings and sequences correspondingly. However there should be more of such characters, not only `-` and `?`. There is also `@` at least. See http://www.yaml.org/spec/1.2/spec.html#ns-plain-first%28c%29

isaacs:
@dervus Simplified the confusing var name, and fixed the style issues you pointed out.  Should be good to go.

dervus:
The only magic here is the fact that we just used the simplest and dumbiest solution for quickly solve the problem and go on. :D Here I'll duplicate the link there you can see how it should be handled: http://www.yaml.org/spec/1.2/spec.html#ns-plain-first%28c%29

lib/js-yaml/dumper.js
isaacs: I think `@` and the other chars will trigger `simple = false`, no matter where they appear in the string.  It seems that only `-` and `?` are ok in a plain scalar as long as they don't appear at the start.

In general, I tried to only add valid folding/literal scalar dumping styles, and not change anything about what was allowed in plain unquoted scalars.

isaacs:
@dervus Ah, ok, that makes a lot of sense.

In that case, yes, there is still an opportunity to do more careful string examination here, if there is a goal to avoid quoting in more cases that could potentially be dumped as plain scalars.  For example, `"a!b"` should be a valid scalar, but as of this pull req will be quoted.

I think we can leave that for the next adventure.  The behavior here is not *incorrect*, and an overall improvement from what was there before.

puzrin:
@dervus , merge?

dervus:
Yeah, there is no much need for making strings plain by any cost. This PR is about block scalars anyway. So thanks for contribution!

dervus:
@isaacs Sorry, I have mislead you about two lines between top-level file statements. Seems like we dropped this style, and I forgot about it. I've corrected that.

puzrin:
I've finished other changesets, but suggest to wait 1-2 days before release, if not urgent. May be someone remember more breaking/pending changes.

puzrin:
Released 3.3.0. Thanks again for your work!

PR #186, opened 2015-04-29
update custom types example

rockitbaby:
tiny update to make to playing with the custom types example more fun.
it does not log to console anymore when module is required


dervus:
Thanks.

PR #188, opened 2015-05-08
Sort keys so that the output is deterministic

rjmunro:
I am dumping data that is generated two different ways, and I want to make sure I get the same result by comparing the output with a diff command. Unfortunately, js-yaml was returning the data in a different order, so my diff was reporting lots of changes that weren't real.

This fixes the problem, ensuring that the output of js-yaml is always the same for the same data.

To do:
- [x] Make this controlled by an option
- [x] Allow passing custom sort function

### DIFF
--- a/test/units/sort-keys.js
+++ b/test/units/sort-keys.js
@@ -1,0 +1,26 @@
+'use strict';
+
+
+var assert = require('assert');
+var yaml   = require('../../');
+
+var sample = { b: 1, a: 2, c: 3 };
+var unsortedExpected = 'b: 1\na: 2\nc: 3\n';
+var simpleExpected = 'a: 2\nb: 1\nc: 3\n';
+var reverseExpected = 'c: 3\nb: 1\na: 2\n';
+
+test('Dumper should sort preserve key insertion order', function () {
+  assert.deepEqual(yaml.safeDump(sample, { sortKeys: false }), unsortedExpected);
+});
+
+test('Dumper should sort keys when sortKeys is true', function () {
+  assert.deepEqual(yaml.safeDump(sample, { sortKeys: true }), simpleExpected);
+});
+
+test('Dumper should sort keys by sortKeys function when specified', function () {
+  assert.deepEqual(yaml.safeDump(sample, {
+    sortKeys: function (a, b) {
+      return a < b ? 1 : a > b ? -1 : 0;
+    }
+  }), reverseExpected);
+});


puzrin:
I don't like this idea. It's not dumper's responsibility to do magic with keys ordering. IMHO you can prepare object prior to dump, or do proxy package

rjmunro:
Preparing the object prior to dumping it is not possible. Objects do not have order
, only arrays have order. Any order they appear to preserve is implementation-dependent. [More info](http://stackoverflow.com/a/5525820/3408)

What do you mean by 'proxy package'?

puzrin:
Object keys order is not guaranteed, you are right. But if you recreate object from scratch, keys will go in creation order in all implementations. I don't like default sorting because it will affect users who expect this behaviour. Probably, this can be useful as not default option.

Under proxy package i mean separate project, that will recreate objects and call js-yaml.

@dervus, @isaacs, what do you think?

rlidwka:
> Objects do not have order, only arrays have order. Any order they appear to preserve is implementation-dependent. [More info](http://stackoverflow.com/a/5525820/3408)

Objects **do** have order called "object enumeration order" which is defined in [section 9.1.12](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys) of the es6 specification. The order is "integer indices in ascending numeric, then strings in property creation order, then symbols in property creation order". And all modern js engines already follow that.

dervus:
IMO it's OK, but only as an option like `sortKeys`, which is disabled by default. Also, it'd be good to allow use of custom sorting function for that.

rjmunro:
@puzrin, @dervus, @isaacs I've added some more commits to make it optional and allow passing a sort function. Are you happy to merge it now?

puzrin:
LGTM

puzrin:
Please, squash into single commit.

dervus:
LGTM

ixti:
Closed via ce0161c9ac004fb5bd5c7a498896bf7a07a68060

ekeren:
Hey, in this current implementation you can't define different sort order based on nested objects 

imagine I want `a' to be before `b` in root element and `b` to be before `a` under `x` element, as far as I can see,  is not possible.



I've already created a solution for this on my local branch and would love to push it. 


There are 3 options here that I see:
1. Adjust the arguments to be nested  - sort key willl be called with the key with nested name 
```javascript
test('Dumper should provide nested keys', function () {
  var nested = { a: 1, b: 2, x: { a: 2, b: 1 } };
  var expected = 'a: 1\nb: 2\nx:\n  b: 1\n  a: 2\n';
  var order = [ 'a', 'b', 'x', 'x.b', 'x.a' ];
  assert.deepEqual(yaml.safeDump(nested, {
    sortKeys: function (a, b) {
      return order.indexOf(a) - order.indexOf(b);
    }
  }), expected);
```
   - Pro: Nice clean and freindly API 
   - Con: Break backward competability 
2. Ad a third argument to sortKey - namespace (array)
```javascript
test('Dumper should provide nested keys', function () {
  var nested = { a: 1, b: 2, x: { a: 2, b: 1 } };
  var expected = 'a: 1\nb: 2\nx:\n  b: 1\n  a: 2\n';
  var rootOrder = [ 'a', 'b', 'x' ];
  var xOrder = [ 'b', 'a' ];
  assert.deepEqual(yaml.safeDump(nested, {
    sortKeys: function (a, b, namespace) {
      if (namespace && namespace.length === 1 && namespace[0] === 'x') { // an ugly way to check if namespace === ['x']
        return xOrder.indexOf(a) - xOrder.indexOf(b);
      }
      return rootOrder.indexOf(a) - rootOrder.indexOf(b);
    }
  }), expected);
```
   - Pro: brackward compatible, doens't add '.' as a special charecter 
   - Con: a sort function with a third parameter - yack! 
3. Do not fix ( i'll fork and use my own implementation) 

I think you should go with option 2, it has an API flaw but it

rjmunro:
@ekeren Another option would be to use a different parameter name than sortKeys. The function you pass to that parameter could have a different signature, similar to option 1, but I think I prefer option 2 as well.

PR #190, opened 2015-05-13
allow astral characters

rlidwka:
According to YAML 1.2 spec, [section 5.1](http://www.yaml.org/spec/1.2/spec.html#id2770814), unicode range #x10000-#x10FFFF is considered printable characters.

However, parser excludes those characters trying to validate for #xD800-#xDFFF. This is fixed in this PR, so well-formed surrogate pairs pass validation, but lone surrogates don't.

It is still not a standard-compliant solution, because YAML seems to allow non-printables in quoted scalars for JSON compatibility, so validating the entire source code against non-printable regexp doesn't work well enough. I added a pending test case for this, suggesting to open another issue ticket about it.

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -23,7 +23,7 @@ var CHOMPING_STRIP = 2;
 var CHOMPING_KEEP  = 3;
 
 
-var PATTERN_NON_PRINTABLE         = /[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\x84\x86-\x9F\uD800-\uDFFF\uFFFE\uFFFF]/;
+var PATTERN_NON_PRINTABLE         = /[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\x84\x86-\x9F\uFFFE\uFFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF]/;
 var PATTERN_NON_ASCII_LINE_BREAKS = /[\x85\u2028\u2029]/;
 var PATTERN_FLOW_INDICATORS       = /[,\[\]\{\}]/;
 var PATTERN_TAG_HANDLE            = /^(?:!|!!|![a-z\-]+!)$/i;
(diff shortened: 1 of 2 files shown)


dervus:
What's about performance of this?

puzrin:
That will be reworked. Wait.

rlidwka:
@dervus, less than ideal probably. I'll try to speed that regexp up. But it is meant to be a temporary solution, because checking non-printables before parsing doesn't really work well (see the point about JSON compatibility, I'll create a separate ticket for it though).

rlidwka:
@puzrin , I created tickets #191, #192 to track those two issues mentioned above

rlidwka:
Joined both regexps into one.

If I didn't screw up, this could even be considered a performance improvement:

```
alex@y:/tmp/js-yaml/benchmark$ git checkout 3b92d922b^
HEAD is now at ad21884... Merge pull request #186 from preciousforever/improvement-custom_types-example
alex@y:/tmp/js-yaml/benchmark$ ./benchmark.js nodeca
Selected samples: (3 of 12)
 > document_nodeca_application
 > document_nodeca_database
 > document_nodeca_logger

Sample: document_nodeca_application.yaml (2055 characters)
 > current x 12,101 ops/sec ±0.57% (92 runs sampled)


Sample: document_nodeca_database.yaml (394 characters)
 > current x 39,270 ops/sec ±0.45% (96 runs sampled)


Sample: document_nodeca_logger.yaml (727 characters)
 > current x 9,787 ops/sec ±0.63% (97 runs sampled)

alex@y:/tmp/js-yaml/benchmark$ git checkout 3b92d922b
HEAD is now at 3b92d92... Process astral characters correctly
alex@y:/tmp/js-yaml/benchmark$ ./benchmark.js nodeca
Selected samples: (3 of 12)
 > document_nodeca_application
 > document_nodeca_database
 > document_nodeca_logger

Sample: document_nodeca_application.yaml (2055 characters)
 > current x 12,389 ops/sec ±1.22% (93 runs sampled)


Sample: document_nodeca_database.yaml (394 characters)
 > current x 40,647 ops/sec ±0.17% (99 runs sampled)


Sample: document_nodeca_logger.yaml (727 characters)
 > current x 12,859 ops/sec ±0.87% (98 runs sampled)
```

dervus:
OK, so is it ready to merge now?

PR #197, opened 2015-07-08
Fixes { "push\npop": env: {} } dumping case.

tcr:
A multiline key as an object key generates invalid YAML. E.g.

```
yaml.safeDump({"push\npop": {env: {}}})
```

yields

```
|-
  push
  pop:
  env: {}
```

This patch corrects this case, but doesn't add tests, etc. I'm happy to do any work to get it mergeable.


puzrin:
Thanks! Could you add tests here https://github.com/nodeca/js-yaml/blob/master/test/units/dump-scalar-styles.js ?

tcr:
@puzrin Added!

PR #200, opened 2015-07-24
Remove unused variables

csun:
I was using uglify to minimize a separate project that uses js-yaml when I noticed a lot of unused variable warnings. After testing, realized that the same warnings come up when performing `make browserify` for js-yaml itself.

I've removed all of the offending variables, and double checked each one to make sure I wasn't deleting anything important. All tests passing and all warnings gone now.


puzrin:
Fixed in master in more accurate way, thanks for report!

PR #205, opened 2015-09-08
fixed broken serialization of duplicated entries in sequences

vogelsgesang:
This pull request fixes a bug in the serialization of duplicated values within arrays. For example, with the following code

``` javascript
var obj = { test: 'canary' };
var array = [ 0, 1 ];
var arrayWithRefs = [ obj, obj, array, array ];
```

`arrayWithRefs` was serialized as

```
- &ref_0test: canary
- *ref0
- &ref_1- 0
  - 1
- *ref1
```

which is incorrect. Line breaks were missing between the reference names (`ref_0`, `ref_1`) and the consecutive characters.

In addition, this PR contains a regression test.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -715,10 +715,6 @@ function writeNode(state, level, object, block, compact, iskey) {
     block = (0 > state.flowLevel || state.flowLevel > level);
   }
 
-  if ((null !== state.tag && '?' !== state.tag) || (2 !== state.indent && level > 0)) {
-    compact = false;
-  }
-
   var objectOrArray = '[object Object]' === type || '[object Array]' === type,
       duplicateIndex,
       duplicate;
@@ -726,6 +722,10 @@ function writeNode(state, level, object, block, compact, iskey) {
   if (objectOrArray) {
     duplicateIndex = state.duplicates.indexOf(object);
     duplicate = duplicateIndex !== -1;
+  }
+
+  if ((null !== state.tag && '?' !== state.tag) || duplicate || (2 !== state.indent && level > 0)) {
+    compact = false;
   }
 
   if (duplicate && state.usedDuplicates[duplicateIndex]) {
@@ -738,7 +738,7 @@ function writeNode(state, level, object, block, compact, iskey) {
       if (block && (0 !== Object.keys(state.dump).length)) {
         writeBlockMapping(state, level, state.dump, compact);
         if (duplicate) {
-          state.dump = '&ref_' + duplicateIndex + (0 === level ? '\n' : '') + state.dump;
+          state.dump = '&ref_' + duplicateIndex + state.dump;
         }
       } else {
         writeFlowMapping(state, level, state.dump);
@@ -750,7 +750,7 @@ function writeNode(state, level, object, block, compact, iskey) {
       if (block && (0 !== state.dump.length)) {
         writeBlockSequence(state, level, state.dump, compact);
         if (duplicate) {
-          state.dump = '&ref_' + duplicateIndex + (0 === level ? '\n' : '') + state.dump;
+          state.dump = '&ref_' + duplicateIndex + state.dump;
         }
       } else {
         writeFlowSequence(state, level, state.dump);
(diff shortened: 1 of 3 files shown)


lib/js-yaml/dumper.js
```diff
@@ -728,6 +724,11 @@ function writeNode(state, level, object, block, compact, iskey) {
     duplicate = duplicateIndex !== -1;
   }
 
+  if ((null !== state.tag && '?' !== state.tag) || duplicate || (2 !== state.indent && level > 0)) {
+    compact = false;
+  }
+
+
```
dervus: Why extra empty line?

lib/js-yaml/dumper.js
```diff
@@ -738,7 +739,7 @@ function writeNode(state, level, object, block, compact, iskey) {
       if (block && (0 !== Object.keys(state.dump).length)) {
         writeBlockMapping(state, level, state.dump, compact);
         if (duplicate) {
-          state.dump = '&ref_' + duplicateIndex + (0 === level ? '\n' : '') + state.dump;
+          state.dump = '&ref_' + duplicateIndex + state.dump;
```
dervus: Is this removed completely? Why? AFAIK this was intended for cases like the following:
``` yaml
a: &ref_1
  foo: bar
  answer: 42
b: *ref_1
```

dervus: But probably I'm mistaken.

lib/js-yaml/dumper.js
vogelsgesang: A mistake... I will fix it. One second.

lib/js-yaml/dumper.js
dervus: Ah, I remembered. It's for anchors on top-level node. So the question stands: Why is this removed?

vogelsgesang: Yes, I removed it completely. I will check if your example still works correctly. In the meantime my reasoning to remove this condition:

The problem with the incorrectly prepended YAML alias names only occurred if `compact` was true. `writeNode` is called with `compact = true` from lines [536](https://github.com/nodeca/js-yaml/blob/0215138ec880c1e0ebc93f5a8c94bb753f5a5456/lib/js-yaml/dumper.js#L536) and from line [625](https://github.com/nodeca/js-yaml/blob/0215138ec880c1e0ebc93f5a8c94bb753f5a5456/lib/js-yaml/dumper.js#L625) and line [830](https://github.com/nodeca/js-yaml/blob/0215138ec880c1e0ebc93f5a8c94bb753f5a5456/lib/js-yaml/dumper.js#L830).

`(0 === level ? '\n' : '')` fixes prepending the alias for top level objects, i.e. for calls from line 830. For example, this snippet ensures that we get

```yaml
&ref_0
- *ref_0
```

instead of

```yaml
&ref_0- *ref_0
```

The two other positions which called `writeNode` with `compact = true` were not handled correctly by this condition.

Since I set `compact` to `false` when an alias is prepended, I think that this condition is not necessary anymore. Without the removal of `(0 === level ? '\n' : '')` we would now actually get

```yaml
&ref_0

- *ref_0
```

(note the extra empty line)

vogelsgesang:
Rebased in order to remove the duplicated empty line I introduced.

lib/js-yaml/dumper.js
vogelsgesang: I just tested your example.

```javascript
var obj = { foo: 'bar', answer: 42 };
var obj2 = { a: obj, b: obj };
console.log(yaml.dump(obj2));
```

is still correctly serialized as

```yaml
a: &ref_0
  foo: bar
  answer: 42
b: *ref_0
```

puzrin:
please remove "make browserify" from PR.

vogelsgesang:
Ok, I removed the "make browserify" commit.

Would be perfect if you could browserify the library then, so that I have a commit to which I can point bower.

puzrin:
Browserification is done for every tagged release. I'll release as soon as @dervus confirm that PR is ok.

lib/js-yaml/dumper.js
dervus: Ah, OK. Thanks for explanation.

dervus:
@vogelsgesang Thanks for contribution!

vogelsgesang:
Thanks for your quick response and for maintaining this library :)

PR #209, opened 2015-10-10
Bump esprima to v2.6.x

shinnn:
npm v3 tries to depdupe the dependencies by default, and keeping dependencies up-to-date helps de-duplication.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,8 @@
     "js-yaml": "bin/js-yaml.js"
   },
   "dependencies": {
-    "argparse": "~1.0.2",
-    "esprima": "~2.2.0"
+    "argparse": "^1.0.2",
+    "esprima": "^2.6.0"
   },
   "devDependencies": {
     "ansi": "*",


package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "argparse": "~1.0.2",
-    "esprima": "~2.2.0"
+    "esprima": "~2.6.0"
```
shinnn: @puzrin How about using `^` operator instead?

puzrin: Ok, let's switch to ^2.6.0. Seems to be ok. Will you update PR?

shinnn: I’ll get right on it.

shinnn: 097666110dc2a657736e1658abc6e193ee2783e3

shinnn:
@puzrin Thanks. Could you publish a new version?

puzrin:
done

shinnn:
:tada:

PR #216, opened 2015-10-28
Add support for typed arrays

koopero:
When trying to serialize object from a physics library, I ran into the following exception:

```
YAMLException: unacceptable kind of an object to dump [object Float32Array]
```

Since typed arrays work almost exactly like regulars Arrays, I added support for them to dumper. A test is included.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -57,6 +57,33 @@ var DEPRECATED_BOOLEANS_SYNTAX = [
   'n', 'N', 'no', 'No', 'NO', 'off', 'Off', 'OFF'
 ];
 
+var IS_TYPE_OBJECT_OR_ARRAY = {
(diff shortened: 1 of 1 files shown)


koopero:
My apologies. This does not pass ci tests for node 0.10.x. Also, there was a typo. It may be that this feature is not testable with your current structure.

I will try to resubmit if I can get it to work with 0.10

dervus:
I think this actually should be implemented via a custom type. Because we should to keep the type. Int32Array isn't just Array, it's Int32Array. Dumped Int32Array should be loaded back as Int32Array.

koopero:
@dervus You are right. My use case was to use yaml to simply trace some objects, without a plan to reload them. I probably went about this the wrong way.

Support for typed arrays is probably still a worthwhile feature, but next time I'll do it right.

Issue #192, opened 2015-05-13
non-printable characters are not supported inside quoted scalars


rlidwka:
YAML standard [section 5.1](http://www.yaml.org/spec/1.2/spec.html#id2770814) explicitly allows non-printables in quoted scalars for JSON compatibility, `js-yaml` doesn't:

```
> require('js-yaml').load('"\ud800"')
JS-YAML: the stream contains non-printable characters at line 1, column 1:
    "���"
    ^
```

```
> JSON.parse('"\ud800"')
'���'
```

PR #219, opened 2015-11-21
Allow non-printable characters inside quoted string literals.

ghost:
Fixed issue #192

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -252,6 +252,8 @@ function captureSegment(state, start, end, checkJson) {
           throwError(state, 'expected valid JSON character');
         }
       }
+    } else if (PATTERN_NON_PRINTABLE.test(_result)) {
+      throwError(state, 'the stream contains non-printable characters');
     }
 
     state.result += _result;
@@ -1521,10 +1523,6 @@ function loadDocuments(input, options) {
   }
 
   var state = new State(input, options);
-
-  if (PATTERN_NON_PRINTABLE.test(state.input)) {
-    throwError(state, 'the stream contains non-printable characters');
-  }
 
   // Use 0 as string terminator. That significantly simplifies bounds check.
   state.input += '\0';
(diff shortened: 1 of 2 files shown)


puzrin:
NB. Speed decrease is ~10% (25% on edge cases). Looks acceptable.

dervus:
Looks good, but

1. Why "JSON-style" input checking is disabled for single quoted scalars? I thought that according to the spec all quoted scalars should follow this exception.

2. Performance probably could be improved by rewriting this regexp to something similar to JSON char check in the same function: https://github.com/nodeca/js-yaml/blob/73c7421f6ee9567c1397631b295ba31be90a3748/lib/js-yaml/loader.js#L245-L255
It was written this way exactly becasue of performance. Manual codepoint-by-codepoint iteration is faster on small inputs than regexp test.

a-rodin:
Disabling of the JSON-style checking for single quoted scalars is a consequence of the following two statements:
a) 5.7. Note that escape sequences are only interpreted in double-quoted scalars
b) 5.1. To ensure readability, non-printable characters should be escaped on output, even inside such scalars.
If there is an non-printable character inside double-quoted scalar, then after parsing and exporting it should be replaced by escape sequence because of b). But it is impossible because of a). So the only way to resolve this problem is to assume that non-printable characters are allowed only inside double-quoted scalars, but not single-quoted ones.

dervus:
You're mixing two independent areas. 5.1 is dumper's concern. Dumper takes JS object and then decides by itself which scalar style to use. 5.7 is out of place at all. It's about `\` escape sequences, no?

I'm just looking for usage of `nb-json` rule, and see that it's base for both `nb-double-char` and `nb-single-char`. See grammer rules in "7.3.1. Double-Quoted Style" and "7.3.2. Single-Quoted Style".

a-rodin:
OK, I've missed the section 7.3.2, so I'll revert the check for single-quoted scalars back.

dervus:
@puzrin, if you think we don't need to try improve the performance, then It's good for merge now,

puzrin:
I doubt that regexp unroll for this case will be faster - too many conditions. Let's leave it for someone else, who has time for experiments.

Issue #222, opened 2015-11-22
Extra new line added after line that longer than 80 symbols


IvanGoncharov:
Test:
```js
var YAML = require('js-yaml');

data = "Exactly 81 symbols in first line. Let's cound finale ones:             9876543210\n test"
console.log(data);
console.log('===============================================================');
data = YAML.safeDump(data);
console.log(data);
console.log('===============================================================');
data = YAML.safeLoad(data);
console.log(data);
console.log('===============================================================');
```
Output:
```
$ node test.js 
Exactly 81 symbols in first line. Let's cound finale ones:             9876543210
 test
===============================================================
>-
Exactly 81 symbols in first line. Let's cound finale
ones:             9876543210

 test

===============================================================
Exactly 81 symbols in first line. Let's cound finale ones:             9876543210

 test

===============================================================
```
So lib add extra newline after string with 80+ symbols.

streamnsight:
set your option for lineWidth to be more than 80 chars.
80 is the default

IvanGoncharov:
@streamnsight Yes I know. It was one of the reason why I added `lineWidth` option.
But still this is workaround and not a solution.

JKillian:
:+1: This behavior surprised me as well.

dervus:
This should be now fixed by #274

PR #223, opened 2015-11-23
Add option to specify max line width.

IvanGoncharov:


puzrin:
Just one question. Is it really necessary to tune width?

IvanGoncharov:
In my case yes, I do some transformations on handwritten YAML files and compare them afterwards.
I work with API descriptions and it most of the cases they have quite long descriptions inside YAML.
For example:  https://github.com/darklynx/swagger-api-collection/blob/master/api/yaml/instagram.yaml#L18
So my intent is to set this values as `-1`.

I took semantic from here: http://stackoverflow.com/questions/17859864/why-does-psych-yaml-interpreter-add-line-breaks-around-80-characters
So I think it useful not only to me :smile:

IvanGoncharov:
Wow it was super fast :+1:

puzrin:
Ok. If you dig closed tickets, you can see that i postponed this feature until someone say "yes i need it" :)

IvanGoncharov:
I found it: https://github.com/nodeca/js-yaml/pull/215#issuecomment-150834932
@puzrin Can you please make release?
This change is blocking me + it allow me to workaround #222

puzrin:
done

IvanGoncharov:
Thanks a lot

PR #224, opened 2015-11-25
Disable tabs indent. Fixes #80.

### LINKED ISSUE #80: tabs indentation is not considered as invalid
Tab indentation will basically result as a non-char when parsing YAML.

For instance:

``` javascript
a:
        b:
                c: yo mamma
```

As tab indented comes out as:

``` bash
js-yaml test.yaml 

{ a: null, b: null, c: 'yo mamma' }
```

It would be safer and easier to debug if the file would not be parseable on tab indent.

ghost:
#80

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -348,6 +348,9 @@ function skipSeparationSpace(state, allowComments, checkIndent) {
         state.lineIndent++;
         ch = state.input.charCodeAt(++state.position);
       }
+      if (ch === 0x09 /* Tab */) {
+        throwError(state, 'tabs indentation is not allowed');
+      }
     } else {
       break;
     }
@@ -798,6 +801,9 @@ function readBlockScalar(state, nodeIndent) {
            (0x20/* Space */ === ch)) {
       state.lineIndent++;
       ch = state.input.charCodeAt(++state.position);
+    }
+    if (ch === 0x09 /* Tab */) {
+      throwError(state, 'tabs indentation in block scalars is not allowed');
     }
 
     if (!detectedIndent && state.lineIndent > textIndent) {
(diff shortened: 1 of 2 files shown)


dervus:
I fear you're turning skipSeparationSpace function into a bigger mess than it already is. The point is — JS-YAML strictly follows (at least tries to) formal grammar definition. Not general statements in the spec. Only formal grammar definition. In this situation we must be able to put a comment at every line of code with a reference to grammar rule in the spec. The problem with skipSeparationSpace is (AFAIK) that it tries to combine different but similar rules under one straight-forward block of code. It was a mistake because it seems like I accidentally mixed two distinct rule into one. These rules as I stated in the very first comment on #80 are:

> [s-l-comments](http://www.yaml.org/spec/1.2/spec.html#s-l-comments) grammar rule is missed, and [s-separate](http://www.yaml.org/spec/1.2/spec.html#s-separate%28n,c%29) is used instead: https://github.com/nodeca/js-yaml/blob/454c7e780cb7b3f9c2fc9f9e03fac1c08c1f2d01/lib/js-yaml/loader.js#L1036

As I also stated there — it's a complicated problem, and I can't just take a look at your solution and say: "Yeah, that's correct". @ingydotnet's words about tabs can't change this either. Yes, I know that tabs are considered as content in YAML, not as indentation. But it is a general statement, and we must follow only grammar rules in the spec. These rules don't say just "Hey, no tabs allowed" as you code assumes if I understood it correctly.

So, what I trying to say is that skipSeparationSpace is messed up in the first place. It implements few rules, and I think nobody can say right now which rules. In this situation I just can't accept any change there without references to rules in the spec. The best solution would be to decompose this function instead to trying to add another hack into it.

By the way, there also should be this test sample:

``` javascript
assert.doesNotThrow(function () { yaml.load('|2\n  \text'); });
```

dervus:
Closed — solution is not correct / inactivity.

Issue #62, opened 2013-02-11
[reader] Reading of properties on implicit block mapping keys.


dervus:
The problem is: we cannot to divide properties of a block mapping from properties of a first _implicit_ entry in that mapping.

So, there will be an error while loading the following document:

``` yaml
block mapping: !!map
  !!str implicit key: some value
```

The reader will see `block mapping:` and interpret it as a block mapping. Then it will try to read properties of that mapping. That is `!!map`. But the reader will not stop and read `!!str` property of the implicit key **as block mapping's** property.

To be clear, there are actually three points of the failure:
- When block collection has a property, the entry list of this collection must be separated from the properties by a line break. (implicit key is not separated from "mapping's" `!!str` property)
- One node cannot have two tags. (`!!map` and `!!str`)
- Resolving of `!!str` tag on a mapping will throw an error.

a-rodin:
It looks like it is only part of the problem and it is also impossible to interpret leading tags in keys, so
```yaml
!!str 16: 32
```
produces exception `YAMLException: end of the stream or a document separator is expected at line`.

brettz9:
any chance this might get revisited? I'm guessing not given the date of the issue, but would sure be nice to support Yarn files in my license-badger project but the Yarn file my yarn install produced gives this kind of error (for `dependencies` here):

```yaml
"@babel/code-frame@7.12.11":
  version "7.12.11"
  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
  integrity sha512-Zt1yodBx1UcyiePMSkWnU4hPqhwq7hGi2nFL1LeA3EUl+q2LQx16MISgJ0+z7dnmgvP9QtIleuETGOiOH1RcIw==
  dependencies:
    "@babel/highlight" "^7.10.4"
```

rlidwka:
@brettz9, this is not a yaml file, so yaml parsers can't parse it. If you check any other yaml parser, they'd fail on it as well.

gregfenton:
I'm still running into what seems like the same issue here?

![image](https://github.com/nodeca/js-yaml/assets/4407559/8e744eb5-8522-4267-a227-f9ca5abcb1db)

Here's the "bad YAML":
```yaml
---
works:
  '1': 11
  !!str 2: 22
  !!str 3: 33
broken:
  !!str 1: 11
  !!str 2: 22
  !!str 3: 33
```

PR #225, opened 2015-11-25
Fixed #62

ghost:


lib/js-yaml/loader.js
```diff
@@ -1100,10 +1102,11 @@ function readTagProperty(state) {
     return false;
   }
 
-  if (null !== state.tag) {
+  if (null !== state.tag && !atLineEnd) {
```
dervus: If I understood this correctly, the you must place this block after `if (atLineEnd)` block instead of disabling it. In current state it would cause a bug allowing to place two or more tags in case of atLineEnd=true.

lib/js-yaml/loader.js
```diff
@@ -1163,6 +1166,19 @@ function readTagProperty(state) {
...
+    while (is_WHITE_SPACE(ch)) {
+      ch = state.input.charCodeAt(++newPosition);
+    }
+
+    if (ch !== 0 && ch !== 0x0A /* LF */ && ch !== 0x0D /* CR */ && ch !== 0x023 /* # */) {
+      state.position = oldPosition;
+      return false;
+    }
+  }
+
```
dervus: OK, but what if we have something like that:
``` yaml
root:
  !!map &1
  !!str key: value
```
where `!!map &1` are properties of the root mapping.

lib/js-yaml/loader.js
```diff
@@ -1243,7 +1259,7 @@ function readAlias(state) {
   return true;
 }
 
-function composeNode(state, parentIndent, nodeContext, allowToSeek, allowCompact) {
+function composeNode(state, parentIndent, nodeContext, allowToSeek, allowCompact, allowPrefixTags) {
```
dervus: If you are adding new function parameter, then **all** calls to that function must explicitly specify a **strictly-typed** argument for it. (i.e. only true or false, never undefined, null, 0, 1 etc)

test/issues/0062.js
```diff
@@ -0,0 +1,22 @@
...
+});
+
+test('Reading top-level mapping with explicit type tag for key', function () {
+  assert.doesNotThrow(function () { yaml.load('!!int 16: 32'); });
+});
+
+test('Reading block mapping with explicit type tag for key', function () {
+  assert.doesNotThrow(function () { yaml.load('block mapping:\n  !!int 16: 32'); });
+});
+
```
dervus: To be honest this test case isn't very useful in my opinion. We shouldn't just check for successful parsing, but also for correct results of the process. I mean we should have there custom types and checks for parsing results, i.e. what tags were applied to what nodes.

For example, in current state this test case will not reveal the bug I mentioned before.

dervus:
I would also add a few general words about the solution itself. It's interesting, but too hacky in my opinion. Source of the #62 defect is some big mistakes made in the core design of JS-YAML since rewrite in 2.x.x. The problem is that readBlockMapping function can read both block mappings AND scalars. I knew how big fuckup it is the day I realized that I have to do it (because we were short at time, JS-YAML rewrite took too long). So, actually back then and until now I think that the *right* way to solve this is to rebuild basic input processing mechanism from being focused on individual nodes to be focused on "pairs". AFAIK there is no such term in the YAML spec, but in terms of JS-YAML non-canonical architecture introduction such new basic parsing element would solve our problems because when parsing pairs of nodes instead of individual nodes we can detect what it is: a mapping or just a scalar without doing into mapping parsing code.

puzrin:
Close - according to comments, need full rework.

Issue #166, opened 2015-02-17
Throw a warning for duplicate keys 


mohsen1:
In [Swagger Editor](https://github.com/swagger-api/swagger-editor) people edit large YAML files and sometimes they mistakenly repeat a key in an object. JSYAML just overrides the first key with second one with no errors or warnings. It would be very helpful if JSYAML was throwing an error or a warning for duplicate keys so I could inform user about the issue.

To be clear, here is an example of duplicate keys:

```yaml
one: 1
one: 2
```
this will be parsed to `{one: 2}`

dervus:
Thanks for pointing this out. Actually it looks like a bug because YAML spec demands that a mapping key must be unique in contrast to JSON (MUST vs SHOULD). However, I dimly remember that there was some reason for current behavior. I must find it out.

PR #228, opened 2015-12-09
Add setting to throw errors when duplicate keys are detected

petey:
It is a really poor experience for my users when their yaml is parsed, and due to the behavior of overriding content of duplicate keys, they do not get what they are expecting. This PR intents to add a flag that will force an error to be thrown when duplicate keys are detected at the same depth of an object. The current default behavior is not changed.

Addresses issue #166


dervus:
Good to merge, but I'm not sure about the option name. Any ideas, @puzrin?

By the way, I think it was done initially without this check because of tests we had. YAML 1.2 spec clearly states that duplicate keys MUST cause an error. The confusion must be came from YAML 1.1 spec, which (AFAIK) does not specify what to do with duplicate keys, and YAML 1.1 implementations (Ruby's for example), which treat duplicate keys just like JS-YAML currently does.

I think we should actually fix our tests top conform YAML 1.2 rules and make the change this pull request adds the default.

petey:
I'm more than happy to make any changes required to have errors thrown as the default behavior. There are a number of tests in the loader test suite that will fail, currently.

puzrin:
No ideas about better names. I'd like to have this mode default on, but that seems to break compatibility with JSON. Can we off it automatically for { } childs ?

puzrin:
My suggestion is to investigate, can we solve this problem without introducing new option - disable dupes for YAML, but enable for JSON. That would be the best solution for users IMHO.

petey:
in JSON, you cannot have duplicate keys at the same level of an object. For example, this is invalid:
```
{
  "one": 1,
  "one": 2
}
```
There should be no difference here between JSON and YAML in this respect.

puzrin:
Som it will be correct to disable dupes completely and bump version to 3.5.0?

@dervus?

dervus:
@puzrin, YAML 1.2 is explicitly incompatible with JSON on this. It's stated in the spec. Actually, JSON [RFC](http://www.ietf.org/rfc/rfc4627.txt) also has a statement for key uniqueness, but it says that keys SHOULD be unique (i.e. it's a recommendation for implementation, not a requirement), and YAML spec says that keys MUST be unique.

In my opinion the best thing to do is to restrict duplicate keys globally by default, and add a `json` option to disable this, and bump the minor version number.

puzrin:
> There should be no difference here between JSON and YAML in this respect.

I've tested:

```bash
$ node -e 'console.log(JSON.parse("{\"a\":1,\"a\":2}"))'
{ a: 2 }
```

Ok, agree that it's better to disable dupes by default and add `json` option.

@petey could you update PR? Also make sure moved tests are working with { json: true }

petey:
I've added the json flag and updated tests.

puzrin:
Squash to single commit please

puzrin:
LGTM after description update.

dervus:
OK, last thing. construct-merge test should not be removed. It's the only test for the merge "type"! This test should also ensure that mapping values are **overridden** without errors if both sides share some keys.

petey:
I still have that test here: https://github.com/petey/js-yaml/blob/strictKeys/test/issues/0166.js#L37-L54

It will fail by default in loader due to duplicate keys.

dervus:
Ah, I see. Actually, you shouldn't do this. Just move  `<<: { ... }` lines to the bottom of each mapping, and that test should be fixed.

petey:
@dervus I tried that.
```
1) Loader construct-merge:
     duplicate key detected: x in "/Users/petey/git/js-yaml/test/samples-common/construct-merge.yml" at line 26, column 7:
      x: 1
```

dervus:
Sorry, I was wrong. I've just refreshed my memories on [!!merge type](http://yaml.org/type/merge.html). This type is implemented as a hack on storeMappingPair-level (there was reasons to) and we need this to be slightly changed to handle both unique key policy and !!merge type behaviour. So, the rules are:

1. If target mapping contains keys `a` and `b` BEFORE `<< : ...` line, then we skip these keys in merge sources.
2. If there are several merge sources (like `<< : [ *CENTER, *BIG ]`), then they override keys of each other from left to right.
3. AFTER `<< : ...` line we must be able to override keys from merged mappings, but NOT keys that were in the mapping before `<< : ...`.

To implement this we need to keep the list of keys merged from other mappings to distinct them from "normal" keys. I think we'll have to keep this list at readBlockMapping/readFlowSequence-level.

petey:
I'm afraid I've never been much good at language parsers, and I can't make heads nor tails of this code. I added a gross hack to make the !!merge test work, but it won't do what you are suggesting it should. If this isn't enough, I'm throwing in the towel. I'm not able to spend any more time trying to figure this out, and you can feel free to close this PR.

dervus:
@petey No problem, I'll fix !!merge by myself on the next weekend then.

dervus:
Resolved.

PR #229, opened 2015-12-11
Add dedup option

devongovett:
Sometimes I don't want to produce references for duplicated objects. This adds an option to avoid that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ options:
 - `sortKeys` _(default: `false`)_ - if `true`, sort keys when dumping YAML. If a
   function, use the function to sort the keys.
 - `lineWidth` _(default: `80`)_ - set max line width.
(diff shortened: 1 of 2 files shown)


dervus:
1. Boolean options must default to false. You should inverse the meaning the of this option.
2. Do not use comparison operators with booleans.
3. Fix code styling violations eslint complains about.

puzrin:
Any ideas about good reversed name for `dedup`?

dervus:
`norefs`

puzrin:
+1

devongovett:
Updated.

puzrin:
Could you squash all to single commit "Added norefs option to dumper"?

devongovett:
Done.

puzrin:
Thanks!

PR #231, opened 2015-12-17
Don't include esprima in browser bundle

### LINKED ISSUE #230: Don't include esprima in the browser bundle
In [function.js](https://github.com/nodeca/js-yaml/blob/master/lib/js-yaml/type/js/function.js) there is a try/catch around a require statement. Browserify will find that require statement and still include esprima in the browser bundle.

Using browserify, you can use the ["browser" field](https://github.com/substack/browserify-handbook#browser-field) like this...
```json
  "browser": {
    "esprima": false
  },
````
...to make `require('esprima')` return an empty object in the browser.

ArtskydJ:
Fixes #230.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     "mocha": "*"
   },
   "browser": {
-    "buffer": false
+    "buffer": false,
+    "esprima": false
   },
   "scripts": {
     "test": "make test"


ArtskydJ:
Closing in favor of #232.

PR #232, opened 2015-12-18
Improve browserify support

### LINKED ISSUE #230: Don't include esprima in the browser bundle

ArtskydJ:
Fixes #230. Supersedes #231.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ browserify:
 	mkdir dist
 	# Browserify
 	( echo -n "/* ${NPM_PACKAGE} ${NPM_VERSION} ${GITHUB_PROJ} */" ; \
-		browserify -r ./ -s jsyaml -x esprima \
+		browserify -r ./ -s jsyaml \
 		) > dist/js-yaml.js
 	# Minify
 	uglifyjs dist/js-yaml.js -c -m \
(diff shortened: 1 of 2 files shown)


puzrin:
Thanks! I plan to make release after #228 fixed or in a week.

TehShrike:
Would you be willing to release this change as a patch version, and release #228 afterward as a feature/minor version?

puzrin:
I'd prefer to increase minor version for this change too.

TehShrike:
Fair enough.

TehShrike:
Any chance this could make it to npm?

puzrin:
That does not work well, because replace module with {} instead if ignoring. It became a problem, when you do really need to use esprima. Replaced your commit with https://github.com/nodeca/js-yaml/commit/6bdb5e79ecd532301a9c2b69c902f35ac72e5455

Issue #235, opened 2016-01-05
round trip error using flow level and string with newline


shaunc:
The following fails:

    yaml.safeLoad(yaml.safeDump({name: '\n'}, {flowLevel: 0}))

producing:
```
YAMLException: missed comma between flow collection entries at line 1, column 8:
    {name: |
           ^
```

The actual output is:

```
> console.log(y.safeDump({name: '\n'}, {flowLevel: 0}))
{name: |
}
```

PR #236, opened 2016-01-05
don't use block-style-only string literals in flow mode

shaunc:
Here is a fix to https://github.com/nodeca/js-yaml/issues/235. I included a test, but I didn't understand your test numbering system.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -240,9 +240,12 @@ function writeScalar(state, object, level, iskey) {
     simple = false;
   }
 
-  // can only use > and | if not wrapped in spaces or is not a key.
-  if (spaceWrap) {
-    simple = false;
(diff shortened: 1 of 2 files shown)


shaunc:
btw ... I tried to install eslint but ... 

$ make lint
Invalid option '--reset' - perhaps you meant '--ext'?
make: *** [lint] Error 1

(I am more familiar with js-hint...)

If you need an explicit version, perhaps better to point the `Makefile` at `node_modules/.bin/eslint` rather than depend on global version?

shaunc:
(slap-on-head -- test numbering system isn't so complex :))

dervus:
Seems good, but could you squish it into single commit?

shaunc:
How's that? I think I got the squash right this time. :)

dervus:
Thanks!

Issue #246, opened 2016-01-25
remove argparse as dependency


Floby:
Hello,
When doing `npm install --production js-yaml`, the installed directory adds up to 3.7Mo including 3Mo of `argparse` which is not used to its full potential.
Consider moving to something either leaner like `minimst` or lighter like `commander`.

Installing for example `nconf-yaml` which only really needs the module, pulls the entire `argparse` + `lodash` (old version, non dedupable) where it is not needed since nconf uses `minimist`.

```
3.0M   node_modules/argparse
92K	   node_modules/commander
100K   node_modules/minimist
740K   node_modules/yargs
```

puzrin:
No. It's a weak reason to change familiar instrument. If you care about size - consider to PR to argparse to minimize dependencies on lodash and so on.

Floby:
Too bad, I was already working on a PR to this =)

gabrielcabola:
@puzrin Do you consider removing argparse@2.0.1?, Since your package pretends to have an MIT license the ArgParse has a Python2.0 license which seems to require a BeOpen license (copyleft):

> BeOpen hereby grants Licensee a non-exclusive, royalty-free, world-wide license to reproduce, analyze, test, perform and/or display publicly, prepare derivative works, distribute, and otherwise use the Software alone or in any derivative version, provided, however, that the BeOpen Python License is retained in the Software, alone or in any derivative version prepared by Licensee.

PR #247, opened 2016-01-25
remove argparse and use commander instead

Floby:
Related to https://github.com/nodeca/js-yaml/issues/246
I gave a spin to try and remove this bloated dependency and use commander which seems to be the lightest option parser around


puzrin:
Sorry, closing as explained in issue.

Floby:
Alright, then using something else =)

puzrin:
No problem.

Issue #248, opened 2016-02-09
expose YAML parsing information [new feature]


ahgittin:
Would be nice if this library could expose intermediate methods for working with the parse tree -- in particular text offsets for model items (including malformed) and comments.

This would allow an API user to edit model content while maintaining a YAML source in synch bidirectionally, with comments and unrelated whitespace preserved.  Also if YAML is malformed in places, the errors could be limited to those sections while nicely handling other sections.  [snakeyaml](snakeyaml.org) is a nice implementation of this (in Java).

I appreciate that this is a significant addition to the scope of this project but I think this is the most complete JS YAML implementation so I'm interested if it's on the roadmap or all or if not whether there might be appetite for it.

puzrin:
That's impossible with current architecture and is not on roadmap, sorry.

ahgittin:
Darn!  :)  Thanks.

ahgittin:
@puzrin a `state.listener` proved a very simple addition sufficient to enable a caller to get the parse tree.  it's low-level but would be really useful for us if you'd consider it!

PR'd in #249.

Just 3 lines of actual code, plus several tests.

puzrin:
I tend to accept it, see comments in PR - some polish needed.

Note, that current architecture was optimized for speed - that's why it may look strange (compared to official yaml parser approach). I think, it can be improved, but experiments require too much of time.

PR #249, opened 2016-02-10
add support for a listener to capture parse events

ahgittin:
This lets a caller build up a parse tree should they need, without impacting anything else.

Inspired by #248 .

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -132,6 +132,7 @@ function State(input, options) {
   this.onWarning = options['onWarning'] || null;
   this.legacy    = options['legacy']    || false;
   this.json      = options['json']      || false;
+  this.listener  = options['listener']  || null;
 
   this.implicitTypes = this.schema.compiledImplicit;
   this.typeMap       = this.schema.compiledTypeMap;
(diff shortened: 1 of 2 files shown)


puzrin:
I don't have principal objection, if you find such hooks useful.

That should not decrease speed, but could you run benchmark for sure? If speed become lower, try to replace `if (state.listener)` with `if (state.listener === null)`

/cc @dervus what do you think?

PS. If you executed tests manually, make sure you run linter too (see travis report)

ahgittin:
Thanks @puzrin .  Delinted.  Benchmarks looked very much the same.  Seemed slightly better with `!== null` -- probably just the OS but it made sense so I've done that instead.

puzrin:
LGTM. @dervus any comments about option name and so on?

dervus:
This change seem to imply public exposure of State object, which I don't like because it would make as to maintain backward comparability for that. I don't object if it'll be non-documented and non-supported feature.

puzrin:
Ok. Let's merge as hidden treasure :)

puzrin:
Released with this PR. Enjoy!

Issue #221, opened 2015-11-22
Extra new line added


IvanGoncharov:
Test:
```js
var YAML = require('js-yaml');

data = "test1\ntest2"
console.log(data);
console.log('===============================================================');
data = YAML.safeDump(data);
console.log(data);
console.log('===============================================================');
data = YAML.safeLoad(data);
console.log(data);
console.log('===============================================================');
```
Output:
```
$ node test.js 
test1
test2
===============================================================
|-
test1
test2

===============================================================
test1
test2

===============================================================
```
So extra new line added in the end of string.

puzrin:
Dupe?

IvanGoncharov:
@puzrin If you ask about #222, then no. It's about trailing new line and not related to 80 symbols.

puzrin:
Probably, that's because of #125 workaround (the same exists in Ruby's parser).

dervus:
Both this and #222 actually caused by combination of two different bugs: one in the loader, and another in the dumper.

Loader bug:
- It doesn't throw an error on block scalar without indentation. Block scalars **cannot** have 0-level indentation.

Dumper bug:
- It doesn't add indentation for top-level block scalar.

dervus:
Dumper part if fixed by #274.

I've created separate issue for the loader bug: https://github.com/nodeca/js-yaml/issues/280

Issue #217, opened 2015-11-11
Only strings should be folded


puzrin:
Currently dumper will fold functions & regexes too. That can cause errors. It seems, folding should be restricted to strings only.

---

@isaacs , since you promoted this feature to dumper & implemented it, i need your opinion. Do you have cases when folding needed for something more than plain strings? Will it affect you if i restrict foltder to strings only?

isaacs:
Restricting folding to strings should be fine, as far as I can see.  If I'd known it would affect other types, I would have added that restriction up front, sorry about the oversight!

puzrin:
No problem. That's not critical since nobody noticed it 6 months after release :)

Issue #183, opened 2015-04-22
[writer] Improve dump style determination for strings


dervus:
In general there are still some cases when we can and probably should use plain scalar style for strings when dumping. So we should not unnecessarily quote URIs for example.

http://www.yaml.org/spec/1.2/spec.html#id2788859

jhartman86:
Is there currently a way to achieve this with the style map options? Another use case would be in a #cloud-init file, for example.
sudo: ['ALL=(ALL) ...']
where currently, there doesn't seem to be a way to output ['ALL...'] without it being wrapped in quotes.

dervus:
Style map is completely unrelated thing. It controls representation style within a node, not node's style itself.

ewaltman:
👍

puzrin:
Close as "endless"

PR #274, opened 2016-03-31
Rewrite scalar dumper for correctness

### LINKED ISSUE #238: Dumper must not add trailing line break at the end
Dumper always adds one trailing line break at the end. It was a feature before we added block scalars support, but now it break dumps in cases, when last object in the stream is `|+` or `>+` scalar.

aepsilon:
Rewrote the scalar dumper with an emphasis on rigor, based on a
careful reading of the YAML 1.2 spec and testing against LibYAML.
### Corrections

Block style chomping accounts for the extra newline
added by subsequent elements or at the end.
Fixes #238.

Top-level block scalars are indented.
Fixes dumper side of #221.

Folded style should be reliable now. The old dumper was not aware of
the subtle rules for more-indented lines.
More-indented lines no longer cause extra newlines to be added.
(The old dumper also had an off-by-one error that dropped
the last character from a long first word on a line.)
Fixes #215, #222. Removes #217 workaround.

Literal and folded styles can represent arbitrary "printable" strings,
including leading/trailing whitespace.
(Hopefully this improves readability for many use cases.)

NB. For the 'construct-string-types' dumper test, the strings are
dumped correctly. The test fails because a loader bug
loses a newline when using the keep "+" chomping indicator.
### Additional improvements

Added test suite covering several former bugs, new edge cases,
and new expected behaviors. All previous applicable tests are included.

Styles are chosen in a well-defined order.
- Plain and single are preferred for single lines under the width limit.
- Folded is preferred when a line is longer than the width limit
  (if enabled) and does not begin with a space.
- Literal is used for all other multi-line "printable" strings.
- Double-quoted is only used when a string is unrepresentable otherwise.

Plain style is relaxed to include more characters and strings.
Progress on #183.

As indent increases, line width decreases down to min(lineWidth, 40).
Setting lineWidth to -1 disables wrapping.
Previously there was unexpected behavior past 40 indent:
narrow lineWidth suddenly jumped to 40,
and -1 (no limit) suddenly enforced a limit of 40.

### DIFF
--- a/test/issues/0217.js
+++ b/test/issues/0217.js
@@ -5,19 +5,107 @@ var assert = require('assert');
 var yaml   = require('../../');
 
 
-test.skip('Javascript functions are not folded', function () {
-  var doc = yaml.load('"key": !<tag:yaml.org,2002:js/function> "function (){ some_function_call(); ' +
-    'and_another_one(); and_another(); and_another(); and_more_and_more(); ' +
-    'until_the_line_become_too_long();}"');
-  var dump = yaml.dump(doc);
-  assert(Math.max.apply(null, dump.split('\n').map(function (str) { return str.length; })) > 80);
+// Simplistic check for folded style header at the end of the first line.
+function isFolded(s) {
+  return s.search(/^[^\n]*>[\-+]?\n/) !== -1;
+}
+
+// Runs one cycle of dump then load. Also checks that dumped result is folded.
+function loadAfterDump(input) {
+  var output = yaml.dump(input);
+  if (!isFolded(output)) {
+    assert.fail(output, '(first line should end with >-, >, or >+)',
+      'Test cannot continue: folded style was expected');
+  }
+  return yaml.load(output);
+}
+
+
+test('Folding Javascript functions preserves content', function () {
+  // Tests loading a function, then tests dumping and loading.
+  function assertFunctionPreserved(functionString, inputs, expectedOutputs, name) {
+    var f = yaml.load('!<tag:yaml.org,2002:js/function> "' + functionString + '"');
+    assert.equal(typeof f, 'function', name + ' should be loaded as a function');
+
+    assert.deepEqual(inputs.map(f), expectedOutputs,
+      name + ' should be loaded correctly');
+
+    assert.deepEqual(inputs.map(loadAfterDump(f)), expectedOutputs,
+      name + ' should be dumped then loaded correctly');
+  }
+
+  // Backslash-escapes double quotes and newlines.
+  function escapeFnString(s) {
+    return s.replace(/"/g, '\\"').replace(/\n/g, '\\n');
+  }
+
+  var fnFactorial = escapeFnString(
+    'function factorial(start) {\n' +
+    '// Non-indented long line to trigger folding: throw new Error("bad fold"); throw new Error("bad fold");\n' +
+    '  var extra_long_string = "try to trick the dumper into creating a syntax error by folding this string";\n' +
+    '  var extra_long_string1 = "try to trick the dumper into creating a syntax error by folding this string";\n' +
+    'var extra_long_string2 = "this long string is fine to fold because it is not more-indented";\n' +
+    'function fac (n) {\n' +
+      'if (n <= 0) return 1; return n * fac(n-1); // here is a long line that can be safely folded\n' +
+    '}\n' +
+    'return fac(start);\n' +
+    '}\n');
+
+  var fnCollatz = escapeFnString(
+    'function collatz(start) {\n' +
+    '  var longString = "another long more-indented string that will cause a syntax error if folded";\n' +
+    'var result = [];\n' +
+    'function go(n) { result.push(n); return (n === 1) ? result : go(n % 2 === 0  ?  n / 2  :  3 * n + 1); }\n' +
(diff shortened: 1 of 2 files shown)


test/issues/0217.js
```diff
@@ -5,7 +5,7 @@ var assert = require('assert');
 var yaml   = require('../../');
 
 
-test('Javascript functions are not folded', function () {
+test.skip('Javascript functions are not folded', function () {
```
puzrin: Why are those tests skipped?

aepsilon: [As @dervus expected](https://github.com/nodeca/js-yaml/pull/215#issuecomment-150790606), the actual problem was the folded style dumper was buggy. It tried to break more-indented lines, which resulted in newlines being added. Folding more-indented lines is not allowed in YAML.

Now that the folder is fixed it should work for all string data regardless of content.
This line tests for a workaround that doesn't fix the root problem (see #222).

Although I think it's better that the root problem is fixed, I can add the workaround back in if desired.

puzrin: I mean, tests were added to check that fns & regexes are not broken. I don't mean tests should stay intact, but those should continue to work. There are no need to disable those cimpletely, for historic reasons.

aepsilon: Naturally, we want to make sure they continue to work.

The dumper doesn't differentiate between tags anymore. The current test demands no folded style, even though the data is dumped correctly. Keeping the test as-is means putting the workaround back in the dumper.

I could change the test to dump then load several examples of functions and regexes.

puzrin: Yes, update tests to reflect your changes. No workaround in dumper needed.

test/units/dump-scalar-styles.js
```diff
@@ -3,79 +3,234 @@
...
+  suite('Folded style', function () {
+    (function () {
+      var content = (function () {
+        var result = '';
+        var i = 1000;
+        for (var para = 1; para <= 7; para++) {
+          result += '\n';
+          // indent paragraphs 3 and 4
+          if (para === 3 || para === 4) {
+            result += ' '.repeat(para);
```
puzrin: `.repeat()` is poorly supported. Needs helper or `Array(para).join(' ')` to fix tests.

Run tests under node 0.10 for sure.

puzrin:
Very impressing. This week surprised me so much - 2 PRs (here & in `pako`), impossible without deep understanding of subject by authors.

@dervus could you take a look?

dervus:
I can take a look on the weekend. Very busy right now, sorry.

aepsilon:
Thanks for the compliment and the quick responses. I've updated the tests.

Regarding the loader bug, it seems to occur whenever the scalar is anything but trailing empty lines:

    > yaml.safeLoad('a: |+\n\n\n')
    { a: '\n\n' }
    > yaml.safeLoad('a: |+\n  \n\n')
    { a: '\n\n' }
    > yaml.safeLoad('a: |+\n b\n\n')
    { a: 'b\n' }

    > yaml.safeLoad('a: >+\n\n\n')
    { a: '\n\n' }
    > yaml.safeLoad('a: >+\n  \n\n')
    { a: '\n\n' }
    > yaml.safeLoad('a: >+\n b\n\n')
    { a: 'b\n' }

lib/js-yaml/dumper.js
```diff
@@ -12,7 +12,7 @@ var _hasOwnProperty = Object.prototype.hasOwnProperty;
 
 var CHAR_TAB                  = 0x09; /* Tab */
 var CHAR_LINE_FEED            = 0x0A; /* LF */
-var CHAR_CARRIAGE_RETURN      = 0x0D; /* CR */
+// var CHAR_CARRIAGE_RETURN      = 0x0D; /* CR */ // unused
```
dervus: No reason to keep it in any form here if it's not in use. It can be just re-added in future if needed.

test/issues/0217.js
```diff
@@ -5,19 +5,107 @@ var assert = require('assert');
...
+    'function go(n) { result.push(n); return (n === 1) ? result : go(n % 2 === 0  ?  n / 2  :  3 * n + 1); }\n' +
+    'return go(start >= 1 ? Math.floor(start) : 1);\n' +
+    '}');
+
+  var fnRot13 = escapeFnString(
+    // single-line function.
+    // note the "{return" is so the line doesn't start with a space.
+    'function rot13(s) {return String.fromCharCode.apply(null, s.split("")' +
+    '.map(function (c) { return ((c.toLowerCase().charCodeAt(0) - 97) + 13) % 26 + 97; })); }'
+  );
```
dervus: I think these functions and constants should be defined at the top-level as two other helper functions.

dervus: I understand the logic though, so it doesn't matter.

dervus:
Also `assert.equal` in tests could be replaced with `assert.strictEqual` for consistency. But I don't really care about it.

Overall that's very good piece of work. Thank you, @aepsilon!

@puzrin, it's good for merge. However, we need to fix that loader bug before releasing this. I'll try to look into it during next week.

puzrin:
https://github.com/nodeca/js-yaml/commit/e0c4e4e3dea07b68bd62022a38944ce3a5b1afb9 - updated asserts use everywhere.

Issue #290, opened 2016-06-03
[writer]Surround string by double quote instead of simple


AlaricCalmette:
I would like to be able to take a json object like this :

```
{
    translation:
        { 
            key: 'this_is_a_key',
            value: 'this is a value'
        }
}
```

and output a yml file like this :

```
- translation :
    key: "this_is_a_key"
    value: "this is a value"
```

I've tried several method but was only able to achieve either

```
- translation :
    key: this_is_a_key
    value: this is a value
```

or 

```
- translation :
    key: '"this_is_a_key"'
    value: '"this is a value"'
```

note that in the latter this is not three quotes but a single quote followed by a double quote

Is there any way to do this with this module?

dervus:
There's no settings for this at the moment. The dumper tries to choose "minimal" possible scalar style to represent each particular string.

Ability to specify preferred scalar styles would be nice addition though. Contributions on this matter are welcomed.

AlaricCalmette:
Okay thanks for the answer, will see if i can contribute.

xavier630:
Any news on this feature?

ratacat:
I'd appreciate this too!

rlidwka:
Fixed by https://github.com/nodeca/js-yaml/commit/7b256d7612a4c29cdf5dfd48a15e01c8dbbb5291 in **dev** branch, gonna be released later.

```js
console.log(require('js-yaml').dump({ key: "this_is_a_key" }, { quotingType: '"', forceQuotes: true }))
// outputs:
// key: "this_is_a_key"
```

PR #291, opened 2016-06-07
Add options to choose prefered scalar style

ychaze:
### Description

Add the ability pass the scalar style (DOUBLE or SINGLE) as option for the dumper.
### Tests

[X] Run all the current test suite by passing the option to the dumper

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -346,11 +346,11 @@ function writeScalar(state, string, level, iskey) {
     var scalarType = chooseScalarStyle(string, singleLineOnly, state.indent, lineWidth, testAmbiguity);
 
     if (state.scalarType === STYLE_DOUBLE) {
-      if (scalarType === STYLE_PLAIN || scalarType === STYLE_SINGLE) {
+      if ((scalarType === STYLE_PLAIN || scalarType === STYLE_SINGLE) && iskey !== true) {
         scalarType = STYLE_DOUBLE;
       }
     } else if (state.scalarType === STYLE_SINGLE) {
-      if (scalarType === STYLE_PLAIN) {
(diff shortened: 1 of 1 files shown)


ychaze:
This should fix the following issue: https://github.com/nodeca/js-yaml/issues/290

AlaricCalmette:
My commit prevent surrounding keys with double or single quote when we specify the scalar type via the parameters of the dumper.

dervus:
1. Option name. First of all there's no such term as "scalar type" in YAML. The name is misleading. I think it should be named `valueFlowScalars` because it's about values, not keys and it's scope is flow scalars only. This name will allow us to conveniently add `keyFlowScalars` and `blockScalars` options in future.
2. The option must accept `plain` (default), `single`, `double` strings as possible values. STYLE_* constants are internal for the dumper. They are not exposed by public API currently, not exposed by this PR, and shouldn't be exposed at all.
3. `if (string.length === 0)` clause was placed after style resolver for absolutely no reason. It should remain were it was. Also I think it's body should be expressed via `?` syntax, not switch.
4. Supplied test doesn't test added feature at all. It tests dumper correctness for the second time while it should ensure that selected scalar styles applied correctly.
5. No documentation entry for new option in README.

Sorry it took me so long to get here. I suppose after so long you may not be interested in rewriting all this, but I have to point out anyway.

puzrin:
Close - no activity/feedback. Let me know if this PR should be reopened or create a new one with updates after @dervus comments.

cula:
Is any step forward here? I want to force add single quote to my string value in yaml. But I cannot find a way neither in Doc nor in Code. I read `chooseScalarStyle` and found no way to make it STYLE_SINGLE.  Is there any solution to it?

PR #296, opened 2016-07-14
add option to specify new line

mathieudutour:


dervus:
1. Option name must conform it's the state field. In this case the option and the field must be named `lineBreaks`.
2. I absolutely don't like idea of allowing to trash dump output by supplying arbitrary line break sequence to the option. Accepted option values must be `unix` (default), `mac`, `windows` strings for `\n`, `\r`, and `\r\n` correspondingly.

mathieudutour:
re 2. my use case was `\n\n` so that the yaml can be read on github (when using front-matter for example). It is still a valid yaml format. But if you really don't like it, then i'll continue using my fork, no biggie

dervus:
Well. I see potential security issue in this. Dunno. @puzrin, your opinion?

mathieudutour:
> potential security issue
I don't understand. Can you explain?

puzrin:
`\n\n` is a strange use case IMHO. If it's not critical, i'd prefer to not extend API for it.

Close?

dervus:
@mathieudutour If this part of parser config comes from insecure source, it can be used to validly alternate dump output. I can't think of any real case, but what's why I said "potential".

For `\n\n` I'd use `yaml.dump(data).replace("\n", "\n\n")`, though I don't really understand the usecase.

mathieudutour:
It's only to dump yaml, not parse it.

And i don't want to replace all new lines, only the one separating fields

dervus:
Equivalence of the dump may be critical.

I may be wrong, but `generateNextLine` is used for all line breaks in the output.

mathieudutour:
the parser already strip the extra new line because it's valid yaml. As long as you specify a valid line break, the parser will be fine.

> I may be wrong, but generateNextLine is used for all line breaks in the output.

not the line breaks in a string

mathieudutour:
> if you really don't like it, then i'll continue using my fork, no biggie

:)

puzrin:
Let's close it as "strange". But i'm open to discussion if new reasons available and so on...

Nepoxx:
Resurrecting an old thread, but a valid use case would be for portability. Windows uses `\r\n` but this package does not respect that (which is fine). Having an option would allow using `require('os').EOL` to make this cross-platform compatible.

puzrin:
1. This is extra dependency, not supported in browser.
2. IMHO it's ugly when result depends on platform.

IMHO if anyone cares about \r\n in dumper, it's more easy to call external replace.

mathieudutour:
It’s not about the extra dependency, it’s about giving the option to use it. 
This PR is just removing a hardcoded value while giving the opportunity to change it. 
I don’t think the fact you don’t understand the external requirements of our use-cases should undermine their validity. 
I don’t see how giving more flexibility while providing sane default could be a bad thing.

puzrin:
@mathieudutour my reply was to previous post, not to you.

I still think this option is specific and can be solved via `.replace()` with minimal effort. The difference is, that API will be more clear.

puzrin:
sorry, i also missunderstood that mesage and thought it suggested to use `require('os').EOL` internally in dumper.

mathieudutour:
No, it’s actually pretty hard to do without the option. It’s not for all the line breaks, only the ones between fields. A `.replace` would also change the line breaks in the values.

puzrin:
Hm... may be i miss something... `\n` can be as code in multiline strings, but it seems ok to replace there too.

Could you give me simple example, when replace `\n` to `\r\n` will break data?

mathieudutour:
One is a value so changing it would change the value while the other is just formatting and wouldn’t matter if you parse it back

puzrin:
Sorry, i don't understand. Could you provide example? I thought, dumper should escape `\n` in strings, and those should not be affected by replace.

Nepoxx:
The solution would be for all OSes to use a consistent newline. In any case, the `replace` suggestion does not support something like this:

```yaml
windowsPath: 'C:\not\going\to\work'
foo: 'bar'
```
If I use a naive replace, this is going to break (there's a `\n` in the path that's not a newline).
Furthermore, I could have something like this:

```yaml
myPublicKey: '-----BEGIN PUBLIC KEY-----\nSomerandomcharacters\n-----END PUBLIC KEY-----'
foo:'bar'
```

In this case, I might not want my public key to have its newline replaces, only the yaml's line endings (mostly for proper displaying in windows).

I know, Windows' newline are annoying, however I can't avoid it.

puzrin:
Output is a string. If it contains `\` + `n` chars sequence, it will not be affected, because only `0x0A` characters will be replaced. Your examples should be ok.

puzrin:
May be i don't know something, and another examples exist. Probably, with multiline strings or binaries. We can discuss those, if you know.

PR #298, opened 2016-09-15
Added travis ci support for modern node versions

amilajack:


puzrin:
We will change it later with other tools update. For example eslint 3.+ does not work with old nodes.

Let's close it now as minor / not needed.

PR #304, opened 2016-10-28
Fix parsing of quotes followed by newlines.

### LINKED ISSUE #303: Loader strips quotes before newlines
`yaml.load("'''foo'' '")` correctly returns the string `"'foo' "`

But if I put a newline in there, the closing quote is lost:
```javascript
> yaml.load("'''foo''\n'")
"'foo "
```

dplepage:
I believe this fixes #303.

The issue is that the loader was setting `captureEnd` incorrectly after absorbing a `''` within a single-quoted string; `captureEnd` would be fixed by the next iteration of the loop on any character other than an end of line.


dervus:
Looks good. Thanks!

PR #310, opened 2016-11-30
Add .gitattributes to force LF checkout

Hirse:
This forces git on Windows to check out the files with the desired line endings even when the recommended global setting of `core.autocrlf` is turned on.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


puzrin:
I don't like changes "just for fun". Will accept when at least one real committer with this problem appears.

Hirse:
Technically, once you merge I become a committer with that problem.
But I understand your decision.

puzrin:
If you decide to fix something serious from tracker, i promise to accept such commit as part of PR.

Hirse:
Deal. 😉

PR #311, opened 2016-11-30
Add information about Node version

Hirse:
Adds information about the supported node versions to the package.json.
Based on the recent change (https://github.com/nodeca/js-yaml/commit/1497178d193fcc380dd30dde0689058cd64896da) that seems to be `>=4.0.0`.


puzrin:
No. It's es5.

Hirse:
What are the supported node versions then?
Why are all <4 no longer tested?

puzrin:
That's only test-related reasons. New eslint requires node 4+, and some tweaks for es5 config required to allow optional features.

Code itself is pure es5 and even works in old browsers with es5 shims.

PR #327, opened 2017-02-03
Report "duplicated mapping key" errors at start of block

### LINKED ISSUE #243: Duplicated mapping key error should point to the key
As revealed in #242 the error points to end of key-value pair instead of duplicate key itself, which is confusing.

shockey:
Resolves #243.

Here, we're storing the starting line and position of a block in `readBlockMapping`, then handing that information to `storeMappingPair ` so it can update the state with that information if it encounters a `duplicated key` error related to that block.

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -281,7 +281,7 @@ function mergeMappings(state, destination, source, overridableKeys) {
   }
 }
 
-function storeMappingPair(state, _result, overridableKeys, keyTag, keyNode, valueNode) {
+function storeMappingPair(state, _result, overridableKeys, keyTag, keyNode, valueNode, startLine, startPos) {
   var index, quantity;
 
   keyNode = String(keyNode);
@@ -302,6 +302,8 @@ function storeMappingPair(state, _result, overridableKeys, keyTag, keyNode, valu
     if (!state.json &&
         !_hasOwnProperty.call(overridableKeys, keyNode) &&
         _hasOwnProperty.call(_result, keyNode)) {
+      state.line = startLine || state.line;
+      state.position = startPos || state.position;
       throwError(state, 'duplicated mapping key');
     }
     _result[keyNode] = valueNode;
@@ -949,6 +951,7 @@ function readBlockMapping(state, nodeIndent, flowIndent) {
   var following,
       allowCompact,
       _line,
+      _pos,
       _tag          = state.tag,
       _anchor       = state.anchor,
       _result       = {},
@@ -969,6 +972,7 @@ function readBlockMapping(state, nodeIndent, flowIndent) {
   while (ch !== 0) {
     following = state.input.charCodeAt(state.position + 1);
     _line = state.line; // Save the current line.
+    _pos = state.position;
 
     //
     // Explicit notation case. There are two separate blocks:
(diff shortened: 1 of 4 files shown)


puzrin:
I have some comments:

1. No need to include `dist/*` to PR
2. Tests needed (with duplicated nested keys too)
3. `storeMappingPair` now is called with differen sets of params. Need to check benchmarks.

shockey:
@puzrin I added tests! 😎  

Can you elaborate on what you mean in your third point?

puzrin:
> Can you elaborate on what you mean in your third point?

JIT can deoptimize functions if they are not monomorphic. Here is good reading http://mrale.ph/.

So, it worth to check, if you change did not slowed down anything. Run `./benchmark/benchmark.js`.

If you see that things become bad, try to add `, 0, 0` to end of all `storeMappingPair` calls.

shockey:
Hey @puzrin, thanks for clarifying.

I ran the benchmark tests before and after my commits, here's the results: https://www.diffchecker.com/Y3Idvbal

puzrin:
Looks similar. Difference can be caused by noise.

Can i ask you do one more thing? Could you change other calls to `storeMappingPair(..., 0, 0)` and compare benchmarks again? I don't like such change, because it looks dirty, but speed make sense. Make sure to close other programs and run twice to check if difference is real or just a noise.

PS. Please, squash your commits into one, i think this PR is valuable enougth and my fantasies about speed should not delay merge :)

shockey:
Yes! I'll change the calls to `storeMappingPair(..., null, null)` instead, because the logic for falling back to what's in state needs a falsy value to kick in.

shockey:
Here's the benchmarks where the number of arguments is constant across the library:

https://www.diffchecker.com/8UMWNP3m

shockey:
Also, GitHub now has a way [to squash pull requests when you merge on GitHub](https://github.com/blog/2141-squash-your-commits). Maybe that will help?

Otherwise, let me know and I can rebase my branch 😎

puzrin:
> Here's the benchmarks where the number of arguments is constant

Equal number of arguments is not enougth. Those should be of the same type in all calls, and `typeof null !== typeof 0`.

As alternative, i'd suggest to pass `(..., -1, -1)` and change condition to `(if >= 0)`

puzrin:
You can just squash commits and i'll merge result as is. It's ok for me. I can test benchmarks myself to save your time.

puzrin:
> Otherwise, let me know and I can rebase my branch

I tried to use github's squash somewhere and don't like garbaged comments it does. So, prefer manual rebase. Please, do it, if not difficult :)

shockey:
Done 🎉

shockey:
Do you mind doing an `npm publish` once this gets merged? We'd love to start using this fix ASAP

puzrin:
LOL. Rebased commit message contains lines i wished to avoid :). I mean second one and below (github adds the same on squash). Could you remove those and leave only first one?

> Do you mind doing an `npm publish` once this gets merged? We'd love to start using this fix ASAP

If that's not deadly critical, i'd like to take a pause for 2 days (not more).

shockey:
^ There's a one-line commit message 👍

puzrin:
Published. Thanks for PR!

PR #328, opened 2017-02-03
Changed dependencies to optionalDependencies

Akkuma:
`js-yaml` could be dependency free for my use case, simple yaml and no CLI use, so I updated it to so that `npm i js-yaml --no-optional` can be leverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+3.8.0 / 2017-02-03
+------------------
+
+- Made dependencies optional rather than required
+
 3.7.0 / 2016-11-12
 ------------------
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-yaml",
-  "version": "3.7.0",
+  "version": "3.8.0",
   "description": "YAML 1.2 parser and serializer",
   "keywords": [
     "yaml",


puzrin:
I don't understand how this change can be useful. Could you explain?

Akkuma:
If you do `npm install js-yaml` you get optionalDependencies installed. If I do `npm install js-yaml --no-optional` I get no optionalDependencies installed. I'm using `js-yaml` to parse a yaml file in Buildkite and need to install it before I can run the script. If I use the latter the reduction in dependencies means the script can start quicker as npm is extra slow on CI and the bottleneck for rather than the code.

puzrin:
I tend to reject, because do not agree with provided arguments about speed. Dependencies are not so big and possible profit was not measured. From the other side, there is a possibility to break something by change `dependency` key with other one. I don't like risky experiments for "nothing".

Akkuma:
There is literally no risk if you look at how `optionalDependencies` works. The only risk would be if `npm` did not work correctly. I'll time it for you when I get a chance for the exact difference locally.

puzrin:
Many programs read `package.json` directly. They can miss check for `optionalDependencies`. And as i said - profit is not evident and was not measured. I think, if you care so much about install speed, it worth to look at `yarn`.

Akkuma:
My timing shows ~2.8s `npm i` with it having at point jumped even higher up to 9.8s and 3.8s in two timings and ~1.8s with `--no-optional`. The assertion that programs read `package.json` is true, but you've created a hypothetical scenario for a hypothetical problem that this would cause. You've lamented that I didn't add measurements, but you're doing the same exact thing by creating this hypothetical problem. Lastly, `yarn` is not production ready for many people and cannot be used by myself due to a bug with handling `git+ssh`, which would require changing packages, their dependencies, and those dependencies' dependencies.

puzrin:
> You've lamented that I didn't add measurements, but you're doing the same exact thing by creating this hypothetical problem

The difference is that you do not respond with your reputation for such problems :) . And i already had issues, related to uncareful package.json change. I can't say if problem really happens, but my experience says risk is too high for package with thousands of dependents.

I could schedule such request for 4.0.0 (far future), but need better understanding of this option in real world:

- which well known (popular) packages use it
- does npm have related issues in tracker (there were something suspicious)

PR #332, opened 2017-03-01
Improve Mark display on compact mode

ooflorent:
Before:

```
in "/Users/fc/Development/js-yaml/test/units/mark.txt" at line 1, column 1
```

After:

```
in "/Users/fc/js-yaml/test/units/mark.txt" (1:1)
```

### DIFF
--- a/lib/js-yaml/mark.js
+++ b/lib/js-yaml/mark.js
@@ -59,9 +59,10 @@ Mark.prototype.toString = function toString(compact) {
     where += 'in "' + this.name + '" ';
   }
 
-  where += 'at line ' + (this.line + 1) + ', column ' + (this.column + 1);
-
-  if (!compact) {
+  if (compact) {
+    where += '(' + (this.line + 1) + ':' + (this.column + 1) + ')';
+  } else {
+    where += 'at line ' + (this.line + 1) + ', column ' + (this.column + 1);
     snippet = this.getSnippet();
 
     if (snippet) {


puzrin:
1. This can be a breaking change, if someone analyze message.
2. I don't like format (x:y) - message become unclear for users.

Could you explain reasons for this PR?

ooflorent:
1. Sure. But is it a big deal? We could say the same thing if a sentence would have been reworded.
2. This is a format adopted by `babel`

The reason behind this change is cosmetic. I have built a tool to format YAML files and errors are handled this way:

```js
try {
  data = yaml.load(contents)
} catch (e) {
  const line = e.mark.line + 1
  const column = e.mark.column + 1
  console.error(`${e.name}: ${e.reason} (${line}:${column})\n`)
  console.error(codeFrame(contents, line, column))
  process.exit(1)
}
```

yml_fmt

The goal is to be able to switch to compact mode to prevent the code frame from being displayed but keeping the mark concise:

* Using `YAMLException#toString()` dumps code 👎 
* Using `YAMLException#toString(true)` hides location 👎 

Modifying `Mark#toString(true)` to be less verbose and adding it to `YAMLException`'s compact output could be a good tradeoff between debugging info and length.

puzrin:
Let's keep this open, to revisit on major version bump.

rlidwka:
Fixed in https://github.com/nodeca/js-yaml/commit/a58d8bc7e1cbdd5a73a67ba41d4c51911efa1aab (together with stylus-like exception formatting).

parkerthierry276-code (approved):

PR #335, opened 2017-03-31
Don't try to parse as an integer when beginning with an underscore

drcapulet:


puzrin:
I did more accurate fix and released it. Thanks for your help!

PR #347, opened 2017-05-12
Return the documents created by loadAll / safeLoadAll

clumsyjedi:
This change makes it possible for calling code to obtain the parsed
documents as an array, or to provide an iterator.


puzrin:
It worth to fix existing tests & add tests for new signatures

clumsyjedi:
I will recreate once done

PR #350, opened 2017-05-24
return parse docs from loadAll

clumsyjedi:


puzrin:
I've commited a more correct and backward-compatible implementation.

PR #353, opened 2017-06-09
Add CDNJS version badge in README.md

sufuf3:
This badge will show the version on CDNJS!


puzrin:
This label does not add value, IMHO.

sufuf3:
OK!
And let me give some information about CDNJS as the followings:
CDNJS is one of the most famous free and public web front-end CDN services which is used by ~1,143,000 websites worldwide.
If people want to use this library, and they want their service faster, they can use Web CDN.
If they know CDNJS host this repository from `README.md`, this will help people when they develop the website.

Thank you!

puzrin:
@sufuf3 i know about CDNJS and have nothing against it. But version info is already available from npm badge, and it's in sync with all CDNs.

PR #358, opened 2017-07-05
Add option to condense output in flow sequences

### LINKED ISSUE #346: Option to omit space between array elements
We like to encode objects and arrays to query parameters using YAML. The parser is able to parse this correctly:

image

But unfortunately, the dumper always adds unneeded spaces between array elements:

image

We don't encode `[]` for prettiness and browsers can handle them fine (PHP has used them forever), but spaces of course do need to be encoded, which is not very pretty.

Here are the options I am currently using to output single line YAML:
```js
yaml.safeDump(value, { flowLevel: 0, indent: 0, lineWidth: Infinity, noCompatMode: true, sortKeys: true })
```
It would be great to have an option for omitting these non-required spaces / emitting "minimized" YAML.

eseliger:
As described in #346, this is useful for using yaml as query parameters in URLs, for example.

Closes #346

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ options:
 - `noRefs` _(default: `false`)_ - if `true`, don't convert duplicate objects into references
 - `noCompatMode` _(default: `false`)_ - if `true` don't try to be compatible with older
   yaml versions. Currently: don't quote "yes", "no" and so on, as required for YAML 1.1
-- `condense` _(default: `false`)_ - if `true` flow sequences will be condensed, omitting the space between `key: value` or `a, b`. Eg. `'[a,b]'` or `{a:{b:c}}`. Can be useful when using yaml for pretty URL query params as spaces are %-encoded.
+- `condenseFlow` _(default: `false`)_ - if `true` flow sequences will be condensed, omitting the space between `key: value` or `a, b`. Eg. `'[a,b]'` or `{a:{b:c}}`. Can be useful when using yaml for pretty URL query params as spaces are %-encoded.
 
 The following table show availlable styles (e.g. "canonical",
 "binary"...) available for each tag (.e.g. !!null, !!int ...). Yaml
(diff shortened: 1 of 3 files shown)


puzrin:
Ok in general, but `condense` option name seems not enougth clear. Any ideas how to improve?

puzrin:
`condensedFlow` (not sure)?

eseliger:
Already had that, too, thought it was too long. But you're right, condense is not very clear. 
If you agree, I could change it to `condenseFlow`. I think the name is ok, it does exactly that, condensing flow sequences, what do you think?

README.md
```diff
@@ -182,6 +182,7 @@ options:
 - `noRefs` _(default: `false`)_ - if `true`, don't convert duplicate objects into references
 - `noCompatMode` _(default: `false`)_ - if `true` don't try to be compatible with older
   yaml versions. Currently: don't quote "yes", "no" and so on, as required for YAML 1.1
+- `condense` _(default: `false`)_ - if `true` flow mode sequences will be condensed, omitting the space between `key: value` or `a, b`. Eg. `'[a,b]'` or `{a:{b:c}}`.
```
puzrin: It's not clear for newbies what is this option for (in real world). May be, add something "may be useful to create compact URL-s"

puzrin:
I have no personal preferences and english is not my native language. So, i'd suggest you to take timeout for 1 day, may be you will get better ideas. This PR will be accepted, but i'd like it to be ideal if possible :)

eseliger:
I'll sleep one night over it, leaving you a comment tomorrow with my thoughts on this :)

eseliger:
Hey, thought about it.

In my opinion, `condenseFlow` describes pretty concise, what this flag does. The only thing this flag changes is omitting the spaces between array entries and and key, value pairs in objects, when in **flow** mode. So, in fact, it does **condense** the output to a minimum, where either arrays or objects are flowing. In conclusion, I think the combination of **condense** and **flow** to `condenseFlow` is probably the most fitting description for this flag. It __condenses items in flow mode__. 

Your thoughts on that? :-)

eseliger:
Great! Could you release this change on npm as well? :) 
That would be awesome!

puzrin:
On weekend ok? Need to rework pending PR

eseliger:
No problem, take the time you need!

puzrin:
published 3.9.0

PR #360, opened 2017-07-30
Ensure stack is present for custom errors (fixes #351)

### LINKED ISSUE #351: CLI no longer provides location of error
The CLI behaviour has changed between node `v6` and `v7`.

Node `v6` with trace (`-t`) option - **Error and location**:
```
$ `npm bin`/js-yaml -t src/index.md
YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key at line 7, column 12:
    description: "Enterprise Content Management ... 
               ^
    at generateError (...
```

Node `v7` with trace (`-t`) option - **No useful information**:
```
$ `npm bin`/js-yaml -t src/index.md
Error
    at generateError (...
```

Node `v7` with compact (`-c`) option - **Error only**:
```
`npm bin`/js-yaml -c src/index.md
YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key
```

I assume this is due to Node no longer providing the same level of detail for errors?
Is there a way to get the location of the error again in `v7`?

not-an-aardvark:
Starting in Node 7, stack traces are computed eagerly when using Error.captureStackTrace ([link](https://chromium.googlesource.com/v8/v8.git/+/4feafee9d9741259e8ec3882fb61935a235ecc54)). As a result, the `message` and `name` properties of an error will not be included in the `stack` property if they are set after calling `Error.captureStackTrace`. As a workaround, set the properties before calling `Error.captureStackTrace`.

Note that this change in Node was reverted in V8 a few days ago ([link](https://chromium.googlesource.com/v8/v8.git/+/989d7b96f8352b502e2ede62e0b105e143d03837)), but it will probably be awhile until that fix appears in a Node release.

Fixes #351

### DIFF
--- a/lib/js-yaml/exception.js
+++ b/lib/js-yaml/exception.js
@@ -6,6 +6,11 @@ function YAMLException(reason, mark) {
   // Super constructor
   Error.call(this);
 
+  this.name = 'YAMLException';
+  this.reason = reason;
+  this.mark = mark;
+  this.message = (this.reason || '(unknown reason)') + (this.mark ? ' ' + this.mark.toString() : '');
+
   // Include stack trace in error object
   if (Error.captureStackTrace) {
     // Chrome and NodeJS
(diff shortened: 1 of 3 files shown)


puzrin:
Thank you! Released 3.9.1

PR #363, opened 2017-08-14
Add `metaKey` option to loader

koczkatamas:
If metaKey option is set in the loader options, then it saves start line and start position of each field under this key into the resulting object.

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -297,7 +297,7 @@ function storeMappingPair(state, _result, overridableKeys, keyTag, keyNode, valu
       _result[state.metaKey] = {};
     }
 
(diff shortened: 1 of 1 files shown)


puzrin:
Looks strange. I don't understand why this needed here. If you  wish to process positions, it looks logical to have intermediate AST implementation, but that's s bit different kind of parser.

koczkatamas:
The result object of the YAML parsing is processed by a library which can throw errors containing the path of erroneous property like "Invalid value found at /path/to/property/which/caused/the/error" and I'd like to show the corresponding line in the text editor.

I think I won't need AST. The line information will be enough to make this possible.

koczkatamas:
Actually I think you're right. I think it's better to go with an AST-based solution.

PR #369, opened 2017-09-05
Combine surrogate pairs into one escape sequence when dumping.

### LINKED ISSUE #368: Support encoding emojis without encoding as surrogate pairs
Currently, if an emoji is in a YAML string, and run through `safeDump`, it will be converted into escaped surrogate pairs (i.e. `thiskey: "😀"` is dumped as `thiskey: "\uD83D\uDE00"`). This is caused by [lib/js-yaml/dumper.js#L468](https://github.com/nodeca/js-yaml/blob/bd32506/lib/js-yaml/dumper.js#L468), as using JS `charCodeAt` returns surrogate pairs by default (see https://mathiasbynens.be/notes/javascript-unicode). Would you be willing to at least make this configurable, so that we could choose whether to convert into surrogate pairs or write the emojis directly? If you want a PR, I can help with that as well, just wanted to see what your thoughts were.

tech4him1:
When an astral character is encountered while dumping to a double-quoted string, save it as an 32-bit escaped Unicode sequence instead of as a surrogate pair (`\U0001F600` instead of `\uD83D\uDE00`).

Closes #368.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -461,11 +461,21 @@ function foldLine(line, width) {
 // Escapes a double-quoted string.
 function escapeString(string) {
   var result = '';
-  var char;
+  var char, nextChar;
   var escapeSeq;
 
   for (var i = 0; i < string.length; i++) {
     char = string.charCodeAt(i);
+    // Check for surrogate pairs (reference Unicode 3.0 section "3.7 Surrogates").
+    if (char >= 0xD800 && char <= 0xDBFF/* high surrogate */) {
+      nextChar = string.charCodeAt(i + 1);
+      if (nextChar >= 0xDC00 && nextChar <= 0xDFFF/* low surrogate */) {
+        // Combine the surrogate pair and store it escaped.
+        result += encodeHex((char - 0xD800) * 0x400 + nextChar - 0xDC00 + 0x10000);
+        // Advance index one extra since we already used that char here.
+        i++; continue;
+      }
+    }
     escapeSeq = ESCAPE_SEQUENCES[char];
     result += !escapeSeq && isPrintable(char)
       ? string[i]


puzrin:
Please don't forget to provide proofs that such encoding is valid.

tech4him1:
@puzrin I'm sorry, are you wanting proof that it is valid according to the spec, or that this code produces valid output?

puzrin:
that it is valid according to the yaml spec

puzrin:
Alternative could be to leave surrogate unencoded. Then after string will be saved in utf-8 (in file, for example), characters will be native.

But this also requires investigation that such thing is allowed in yaml.

tech4him1:
Encoding astral characters as 32-bit escape sequences (like `\U0001F600`):
- This method is provided in the YAML spec here: http://www.yaml.org/spec/1.2/spec.html#ns-esc-32-bit
- Also, this is the method that PyYAML uses:
  ```py
  >>> import yaml
  >>> print(yaml.dump({'testkey': "🐣 🌼 🌷 🌱 🌳 🍃 🌈"}))
  testkey: "\U0001F423 \U0001F33C \U0001F337 \U0001F331 \U0001F333 \U0001F343 \U0001F308"
  ```
- It is also the method `go-yaml` and the [reference parser](http://ben-kiki.org/ypaste/) uses.
- Here is a reply I received from Leon Timmermans on the YAML core mailing list:
  > Javascript (and a few other languages with UTF-16 implementation details
leaking out) has a tendency to treat such characters as two surrogates
("\uD83D\uDCA9"), instead of as a single character ("\U0001F4A9"). Quite
frankly I think this is unhelpful and wrong, but JSON actually made it a
standard -_-.

puzrin:
Exactly. References are solid. Thank you.

One more question. Could this be a problem with JSON in JS? (i think - no)

tech4him1:
Partially, yes. There are two related things here that are causing this in JS and JSON.

The first is that JavaScript internally uses either UCS-2 or UTF-16 (https://mathiasbynens.be/notes/javascript-encoding). So whenever you use `string.charCodeAt(i)` it will return the surrogate pairs, as that is how they are actually stored in the string. ES6 introduces [`string.codePointAt(i)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt) which takes surrogate pairs into account, but I didn't think you wanted to use ES6 here, so as to keep `js-yaml` backwards compatible.

The second issue is that in JSON, astral characters *are supposed to be* encoded as surrogate pairs (see [JSON spec](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) section "9 String"), as JSON does not have long escape sequences. This does make it confusing to convert between JSON and YAML.

tech4him1:
@puzrin Also, do you think this should be an option, or the default? If an option, do you have an idea what it should be named (maybe `astralsAsSurrogates`)?

puzrin:
I think - default will be good enougth.

tech4him1:
@puzrin Also, the loader should already be able to parse these 32-bit sequences, correct? https://github.com/tech4him1/js-yaml/blob/1393e2539556776899d274cce2cc972b5b9ae69f/lib/js-yaml/loader.js#L114-L119

tech4him1:
This PR should be ready to merge, unless you have any changes that you want made.

puzrin:
published

tech4him1:
Thank you!

rlidwka:
This code was changed in https://github.com/nodeca/js-yaml/commit/0b2e7029a137935279a4d3e3d26cc63cdab8b57d, astral characters are no longer escaped, see discussion in https://github.com/nodeca/js-yaml/issues/587.

PR #378, opened 2017-10-29
feat: duplicate check for objects

jroehl:
### Changes
Previously the duplicate check implementation was only working for primitive types (`indexOf()`)
- Added `refs` duplicate check for objects

### Affected scripts
- `js-yaml/lib/js-yaml/dumper.js`
- `js-yaml/test/20-dumper.js` (had to switch off refs to make tests work, will add further tests for refs if required)

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -176,6 +176,24 @@ function isWhitespace(c) {
   return c === CHAR_SPACE || c === CHAR_TAB;
 }
 
+// Returns true if the two objects are equal
+function isEqual(a, b) {
+  var aProps = Object.getOwnPropertyNames(a);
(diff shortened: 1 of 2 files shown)


puzrin:
Existing implementation works as expected - detects references to the same object. Your change tries to do magic and i don't like this.

jroehl:
Ok, so its only supposed to work for primitive types?

puzrin:
In JS non-primitive objects are equal if references point to the same object. `Array.indexOf()` is ok for references tracking. If you dump content with duplicated (by reference) leaves, that will be detected.

puzrin:
Suggested method of objects compare is:

1. not universal (does nor compare classes, for example)
2. very slow

I think, if you wish to have such specific optimization, it's better to preprocess your data before dumper.

PR #381, opened 2017-11-21
Options not getting passed through safeLoadAll when not specifying iterator

nmorsman:
Options doesn't get passed through correctly when using `safeLoadAll()` and no iterator.

Example:
```js
var data = '---\n' +
'  test: 123\n' +
'  test: 456';

// Works
yaml.loadAll(data, null, { json: true });

// Exception
yaml.safeLoadAll(data, null, { json: true });
```


zjffun:
I'm sorry I didn't check prs before submitting #547 . Oh, this bug existed for so long. 

But according to my understanding the api wants to express that those two methods treat the second parameter as options when omit iterator.

```
safeLoadAll (string [, iterator] [, options ])
loadAll (string [, iterator] [ , options ])
```

PR #392, opened 2018-01-23
add replacer code similar to JSON replacer and (configurable) toJSON

sjpt:
I apologize if I haven't created this request correctly; still very unused to git.
I have changed the base file lib/js-yaml/dumper,js, but not built any dist/js-yaml.js or similar files.
Also, there are just two test cases which are for the moment included inline in dumper.js.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -124,6 +124,8 @@ function State(options) {
 
   this.duplicates = [];
   this.usedDuplicates = null;
+  this.replacer = options.replacer;  // sjpt
+  this.localReplacer = options.localReplacer;  // sjpt
 }
 
 // Indents every line in a string. Empty lines (\n only) are not indented.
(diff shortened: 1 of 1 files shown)


lib/js-yaml/dumper.js
```diff
@@ -682,7 +684,31 @@ function detectType(state, object, explicit) {
...
+  failing test case, because keys get processed through this as well as values
+    obj = {x:3, y:7, zzz: {q:3, toJSON: function(m) {return this.q || 7}}};
+    repl = (k,v)=> v === 'y' ? k + 993 : v
+    yyy = jsyaml.safeDump(obj, {replacer: repl , localReplacer: 'toJSON', skipInvalid: true})
+    jjj = JSON.stringify(obj, repl)
+    console.log('yyy => ', yyy); console.log('jjj => ', jjj)
+  ***/
+  let object = oobject; // keep original for debug
+  if (typeof object === 'object' && object[state.localReplacer]) {
+    object = object[state.localReplacer].apply(object);
```
graphicore: This is not called  correctly:

> JSON.stringify() calls toJSON with one parameter:
>
> * if this object is a property value, the property name
> * if it is in an array, the index in the array, as a string
> * an empty string if JSON.stringify() was directly called on this object


 do it like this:

```js
object_ = object[state.localReplacer](key || '');
```

**NOTE:** we have a problem here, `writeNode` is not aware of the key it is writing.

Also, not each invocation of  `writeNode` writes a value, some invocations just write keys, see the `isKey` parameter. If `isKey` is true, we should never even consider calling any replacer.

lib/js-yaml/dumper.js
```diff
@@ -682,7 +684,31 @@ function detectType(state, object, explicit) {
...
+  }
+  if (state.replacer) {
```
graphicore: that's not sufficient to match https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter

lib/js-yaml/dumper.js
```diff
@@ -682,7 +684,31 @@ function detectType(state, object, explicit) {
...
+
+  failing test case, because keys get processed through this as well as values
+    obj = {x:3, y:7, zzz: {q:3, toJSON: function(m) {return this.q || 7}}};
+    repl = (k,v)=> v === 'y' ? k + 993 : v
+    yyy = jsyaml.safeDump(obj, {replacer: repl , localReplacer: 'toJSON', skipInvalid: true})
+    jjj = JSON.stringify(obj, repl)
+    console.log('yyy => ', yyy); console.log('jjj => ', jjj)
+  ***/
+  let object = oobject; // keep original for debug
+  if (typeof object === 'object' && object[state.localReplacer]) {
```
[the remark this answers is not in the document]
graphicore: This:

```js
&& object[state.localReplacer]
```

must be rather:

```js
&& Object.prototype.hasOwnProperty.call(object, state.localReplacer)
```

rlidwka:
implemented slightly differently in https://github.com/nodeca/js-yaml/commit/359b264c0fbadfae86caff91879bf40e902df690

PR #397, opened 2018-02-04
fix: don't import esprima in webpack builds

eseliger:
This makes webpack also not discover the import, so esprima won't be included in these builds either. Reduces bundle size a lot, let me know if you want any additional changes :)

### DIFF
--- a/lib/js-yaml/type/js/function.js
+++ b/lib/js-yaml/type/js/function.js
@@ -11,8 +11,8 @@ var esprima;
 //
 try {
   // workaround to exclude package from browserify list.
-  var _require = require;
-  esprima = _require('esprima');
+  // eslint-disable-next-line no-eval
+  esprima = eval('require')('esprima');
 } catch (_) {
   /*global window */
   if (typeof window !== 'undefined') esprima = window.esprima;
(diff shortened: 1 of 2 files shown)


lib/js-yaml/type/js/function.js
```diff
@@ -11,8 +11,8 @@ var esprima;
 //
 try {
   // workaround to exclude package from browserify list.
-  var _require = require;
-  esprima = _require('esprima');
+  // eslint-disable-next-line no-eval
+  esprima = eval('require')('esprima');
```
puzrin: I'd like to avoid such `eval` use.

eseliger: I know its not perfect.. I tried many other things but apparently, webpack is quite intelligent. Another way would be to use the `IgnorePlugin` webpack provides, which is less obvious. If you want it to be so, I would be down to write a note in the readme that this should be considered to keep bundle size low. It was more of an accident that I found it in our bundle and I'd like to save others the time for finding out :)

puzrin:
Let's stay with readme comment for now.

puzrin:
Close. Comment in readme welcome.

PR #401, opened 2018-03-01
char -> character

gseyffert:
This is minor, but when using this library with the Closure compiler, `char` is a reserved keyword and it will fail to compile. Changing these occurrences of `char` to `character` fixes things for me.
Clearly most people aren't affected by this, as I can't find anything in Issues, but I thought a PR submission would be worth it since we are looking to use this library in conjunction with the Closure compiler in a production environment, and not having to maintain a custom fork/package to get our builds to work would be ideal.
Thanks for all the work you've put into this library!


puzrin:
1. I don't like this such rename.
2. Reason is suspicious. I don't beleive it can't be solved in another way.

puzrin:
I don't beleive that CS optimizer has limitations for processed JS. That's not normal. I'd suggest to investigate situation with your instrument, instead of spreading kludges worldwide.

PR #404, opened 2018-03-15
Check for leading newlines when determining if block indentation indicator is needed

### LINKED ISSUE #403: Incorrect dump of block scalar with leading newlines and spaces
The following code dumps invalid YAML:

```js
console.log(yaml.safeDump({ str: '\n    a\nb' }))
```

Invalid output due to indentation:

```yaml
str: |-

      a
  b
```

Leading spaces are handled in general by an indentation indicator (below), but leading newlines seem to thwart this (above):

```js
console.log(yaml.safeDump({ str: '    a\nb' }))
```

Valid output:

```yaml
str: |2-
      a
  b
```

The results above are from version 3.11.0.

trevorr:
Fixes #403

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -307,7 +307,7 @@ function chooseScalarStyle(string, singleLineOnly, indentPerLevel, lineWidth, te
       ? STYLE_PLAIN : STYLE_SINGLE;
   }
   // Edge case: block indentation indicator can only have one digit.
-  if (needIndentIndicator(string) && indentPerLevel > 9) {
+  if (indentPerLevel > 9 && needIndentIndicator(string)) {
     return STYLE_DOUBLE;
   }
   // At this point we know block styles are valid.


lib/js-yaml/dumper.js
```diff
@@ -301,7 +307,7 @@ function chooseScalarStyle(string, singleLineOnly, indentPerLevel, lineWidth, te
       ? STYLE_PLAIN : STYLE_SINGLE;
   }
   // Edge case: block indentation indicator can only have one digit.
-  if (string[0] === ' ' && indentPerLevel > 9) {
+  if (needIndentIndicator(string) && indentPerLevel > 9) {
```
aepsilon: Looking back, an optimization would be to short-circuit by first comparing `indentPerLevel > 9`, because it's almost always something like 2, 4, or 8. Likely not a big deal though.

trevorr: Makes sense. I'm happy to change it, or you can.

puzrin:
Thanks!

Issue #389, opened 2017-12-27
Support for ES6 in `js/function` YAML type


julien-c:
```yaml
function: !!js/function >
  () => 'Wow! JS-YAML Rocks!';
```

puzrin:
Looks reasonable. PR will be accepted.

kivlor:
I've had a crack at adding support for ES6 arrow functions, see - https://github.com/nodeca/js-yaml/pull/393

For the moment, dumped arrow functions are just 'standard' functions. JS doesn't seem to have an `ArrowFunction` class so I couldn't see a way to dynamically create them.

I could use `eval`? (joking, not joking...)

aengl:
I tried to support arrow functions without block statements, feedback welcome:

https://github.com/nodeca/js-yaml/pull/421

PR #421, opened 2018-05-17
Support arrow functions without a block statement

aengl:
#393 supports arrow functions, but will fail when using arrow functions without block statements, like the one mentioned by @julien-c in #389:

```
function: !!js/function >
  () => 'Wow! JS-YAML Rocks!';
```

Like @kivlor pointed out already, there's no constructor for `ArrowFunction`, so this is a bit of a workaround. Personally, I'd rather see this resolved with a `safe-eval`.

### DIFF
--- a/lib/js-yaml/type/js/function.js
+++ b/lib/js-yaml/type/js/function.js
@@ -65,8 +65,14 @@ function constructJavascriptFunction(data) {
 
   // Esprima's ranges include the first '{' and the last '}' characters on
   // function expressions. So cut them out.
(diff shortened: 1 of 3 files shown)


puzrin:
LGTM.

Kick me in a couple of days, i'll merge & publish if no more objections.

aengl:
@puzrin Ping 😃

puzrin:
Sorry for silence. Published new version with your fixes. Thanks for efforts and patience!

dvv:
Converting `({ foo, bar}) => "baz"` eats argument. Please consider to support the case. TIA

PR #453, opened 2018-10-26
Added Dockerfile for 'js-yaml' command line utility

pjzedalis:
Convert yaml to json in real time via shell:

docker run --rm -v $(pwd):/opt serverwarp/js-yaml:3.12.0 js-yaml /opt/test.yaml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM alpine:3.8
+MAINTAINER Phillip Zedalis <phillip@serverwarp.com>
+
+# Get updates
+RUN apk update
+RUN apk upgrade
+
+# Install NodeJS & NPM
+RUN apk add nodejs
+RUN apk add npm
+
+# Install js-yaml
+# https://github.com/nodeca/js-yaml
+RUN npm install -g js-yaml


pjzedalis:
Based on Alpine 3.8

puzrin:
Idea looks interesting. But at first glance:

- https://hub.docker.com/r/glowbal/docker-js-yaml/
- https://hub.docker.com/r/simplealpine/yaml2json/

Are you still sure Dockerfile is really needed in this repo? If yes - could you explain?

puzrin:
Close - timed out. Seems not needed.

pjzedalis:
I made it because I needed it. The first link you provided is way out of date. The second is not the same thing. If you don't need it no problem. :)

puzrin:
I'm very conservative about adding new features into stable packages. That means, for any proposal some questions should be answered:

- is it possible to get result in other way 
- how many people need it
- is it required here or in separate package

So, while i have nothing against your idea in general, it really worth to investigate it more deep:

- are there simillar ready projects on docker hub?
- isn't it better to use pyyaml?
- how to publish/update docker hub project?

If you decide to continue - feel free to open new issue with your thoughts/details and we will discuss those. Thanks for understanding.

PR #454, opened 2018-10-29
Add ability to override quoting style for scalar values

### LINKED ISSUE #290: [writer]Surround string by double quote instead of simple

jpb:
This is a reimplementation of #291 and applies the feedback from the original implementation in https://github.com/nodeca/js-yaml/pull/291#issuecomment-238114106, except in two cases:

- the option name has been changed from `valueFlowScalars` to `scalarQuoteStyle` as the original implementation was (and my use-case is) not limited to flow values
- I chose to not accept `STYLE_PLAIN` as a value. This didn't make sense to me as I feel this option should never "downgrade" a values quoting – js-yaml only uses quotes when necessary to make valid YAML so removing single or double quotes would to make the values "plain" would produce invalid YAML. This is also why `STYLE_PLAIN` will have no effect for double quoted values (as the readme states).

Fixes #290

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ options:
 - `noCompatMode` _(default: `false`)_ - if `true` don't try to be compatible with older
   yaml versions. Currently: don't quote "yes", "no" and so on, as required for YAML 1.1
 - `condenseFlow` _(default: `false`)_ - if `true` flow sequences will be condensed, omitting the space between `a, b`. Eg. `'[a,b]'`, and omitting the space between `key: value` and quoting the key. Eg. `'{"a":b}'` Can be useful when using yaml for pretty URL query params as spaces are %-encoded.
+- `scalarQuoteStyle` _(default: `null`)_ set the quote style for scalar values. Valid values are `"single"` and `"double"`. Double quoted values will not be changed to single quotes.
 
 The following table show availlable styles (e.g. "canonical",
 "binary"...) available for each tag (.e.g. !!null, !!int ...). Yaml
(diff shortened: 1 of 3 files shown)


puzrin:
Implementation is very dirty:

- multiple value types on input
- strings compare in hot path

What should be done:

- use empty string / ' / " as input.
- normalize input value (string) to number, before run.
- something should be done with option name (see comments in prev commit), need to discuss

yihongang:
@jpb are you still interested in continuing with this work?

@puzrin could you explain what do you mean by input here? is it the `null` default value vs the expected string values? with regards to the option name what do you have in mind?

mansona:
I have just tested this and it's working really nicely 🎉 seems like something that would be good to look to merge

nicmosc:
What's the status on this? Really need this feature!  🙏🏻

kevthehermit:
Any movement on this?

SkinnyFatSumo:
Any chance this feature will come through anytime soon? It'd be very helpful

rlidwka:
I've added more extensive support for this feature in https://github.com/nodeca/js-yaml/commit/7b256d7612a4c29cdf5dfd48a15e01c8dbbb5291 (located in **dev** branch currently).

Closing this as outdated. Thanks for posting this PR, helped to find some corner cases I haven't originally thought of.

Issue #432, opened 2018-07-04
list indention style is different from pyyaml


thewilli:
consider the following `pyyaml` code ([try yourself](https://repl.it/repls/TidyInformalAutomatedinformationsystem)):

```python
import yaml

print yaml.dump(
  {
    "foo": [1, 2, 3]
  },
  default_flow_style=False
)
```
which outputs
```yaml
foo:
- 1
- 2
- 3
```` 
and the comparable `js-yaml` counterpart ([try yourself](https://repl.it/repls/ConsiderateSteepProgramminglanguage)):

```js
const yaml = require("js-yaml");

console.log(
  yaml.safeDump({
    foo: [1, 2, 3]
  })
);
```` 
which outputs

```yaml
foo:
  - 1
  - 2
  - 3
```

You can see that `pyyaml` doesn't add spaces to lists (because `-` counts as indention). I know both are valid, but it would be great if we could a similar output as `pyyaml` (especially as this project started as some kind of fork)?

Changing [this line](https://github.com/nodeca/js-yaml/blob/2d1fbed/lib/js-yaml/dumper.js#L738) to

```js
writeBlockSequence(state, level - 1, state.dump, compact);
```

changes the behaviour so it should be possible to add a configuration setting to allow both behaviours.

puzrin:
IMHO this request is specific and there are no big reasons to change output format or add an option. I tend to close it.

thewilli:
> IMHO this request is specific and there are no big reasons to change output format or add an option. I tend to close it.

After doing a small evaluation I'm pretty sure most of (many) frameworks are doing what `pyyaml` does. And shouldn't this library tend to stick to what appears to be standard (i.e. not emitting additional, needless whitespaces)?

puzrin:
- Probably, "most of frameworks" just are just binding to libyaml. There are no standard OR recommendations about better output look.
- "needless whitespaces" is not good argument, because yaml is about readability, not about size optimizations

You request a breaking change without valuable profit (at least, i could not understand it) - i don't like such balance.

puzrin:
"industry convention" is marketing bullshit bingo slang. I'd suggest to focus on technical arguments instead.

What's wrong with "other libraries and software"? If dumper produce valid yaml, why should someone fight with spaces?

thewilli:
> "industry convention" is marketing bullshit bingo slang.

[...]

PR #461, opened 2019-01-04
Add "noArrayIndent" option

jacob-hd:
Addresses issue #432 by adding a `noArrayIndent` option to optionally not add an extra level of indentation to array elements.

When `noArrayIndent` option is set to `false` (or not provided), output is:
```
array:
  - a
  - b
  - c
```

When `noArrayIndent` option is set to `true`, output is:
```
array:
- a
- b
- c
```

This helps avoid diffs when parsing, modifying, and generating valid yaml that does *not* use extra indentation for arrays.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -105,16 +105,17 @@ function encodeHex(character) {
 }
 
 function State(options) {
-  this.schema       = options['schema'] || DEFAULT_FULL_SCHEMA;
-  this.indent       = Math.max(1, (options['indent'] || 2));
-  this.skipInvalid  = options['skipInvalid'] || false;
-  this.flowLevel    = (common.isNothing(options['flowLevel']) ? -1 : options['flowLevel']);
-  this.styleMap     = compileStyleMap(this.schema, options['styles'] || null);
-  this.sortKeys     = options['sortKeys'] || false;
-  this.lineWidth    = options['lineWidth'] || 80;
-  this.noRefs       = options['noRefs'] || false;
-  this.noCompatMode = options['noCompatMode'] || false;
-  this.condenseFlow = options['condenseFlow'] || false;
+  this.schema        = options['schema'] || DEFAULT_FULL_SCHEMA;
+  this.indent        = Math.max(1, (options['indent'] || 2));
+  this.noArrayIndent = options['noArrayIndent'] || false;
+  this.skipInvalid   = options['skipInvalid'] || false;
+  this.flowLevel     = (common.isNothing(options['flowLevel']) ? -1 : options['flowLevel']);
+  this.styleMap      = compileStyleMap(this.schema, options['styles'] || null);
+  this.sortKeys      = options['sortKeys'] || false;
+  this.lineWidth     = options['lineWidth'] || 80;
+  this.noRefs        = options['noRefs'] || false;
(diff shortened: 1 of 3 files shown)


lib/js-yaml/dumper.js
```diff
@@ -107,6 +107,7 @@ function encodeHex(character) {
 function State(options) {
   this.schema       = options['schema'] || DEFAULT_FULL_SCHEMA;
   this.indent       = Math.max(1, (options['indent'] || 2));
+  this.indentArrays = (options['indentArrays'] === undefined) ? true : !!options['indentArrays'];
```
jacob-hd: Defaulting to true to maintain backwards-compatibility.

.gitignore
```diff
@@ -5,3 +5,4 @@ doc
 benchmark/implementations/*
 !/benchmark/implementations/current/
 .idea
+package-lock.json
```
jacob-hd: The package lock should either be committed or ignored. Seeing as it isn't committed, I'm ignoring it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "js-yaml",
-  "version": "3.12.0",
+  "version": "3.13.0",
```
jacob-hd: Not sure if I'm supposed to version bump or not...

.gitignore
puzrin: Don't touch configs if PR is not related

lib/js-yaml/dumper.js
```diff
@@ -107,6 +107,7 @@ function encodeHex(character) {
 function State(options) {
   this.schema       = options['schema'] || DEFAULT_FULL_SCHEMA;
   this.indent       = Math.max(1, (options['indent'] || 2));
+  this.indentArrays = (typeof options['indentArrays'] === 'undefined') ? true : !!options['indentArrays'];
```
puzrin: I'd suggest `noArrayIndent` instead of such logic.

package.json
puzrin: Don't change configs.

puzrin:
Thanks! I know about "squash merges". But it just more convenient for review to to collapse long tail of commits, when only couple of files affected.

puzrin:
@jacob-hd could you take a look at https://github.com/nodeca/js-yaml/issues/468#issuecomment-464847252?

Seems indents become broken if array contains objects

puzrin:
@jacob-hd 

```
node -e "console.log(require('js-yaml').dump({ test: [[1, 2], [3,4]]}, { noArrayIndent: true }))"
```

=>

```
- - 1
- 2
- - 3
- 4
```

One more example of bad behaviour. Seems initial PR has not enougth nested samples.

procaconsul:
Following on from what @puzrin said about the issue with noArrayIndent, the following also does not work:

```
node -e "console.log(require('js-yaml').dump({ test: [1, 2, 3, 4]]}, { noArrayIndent: true }))"
```
=>
```
test:
  - 1
  - 2
  - 3
  - 4
```
vs expected 
```
test:
- 1
- 2
- 3
- 4
```

I suspect that the issue might be resulting from an earlier fix by @diberry (namely https://github.com/diberry/js-yaml/commit/18a871cf69435309598393be86006e39abc9d492), where the level should be >= 0, rather than just > 0, but I am not entirely sure. Just trying to provide a pointer :)

PR #435, opened 2018-08-05
Add "main" field to package.json

pelotom:
This allows WebPack to automatically find the adjacent `js-yaml.min.js` when making an optimized build, which greatly reduces bundle size.

PR #462, opened 2019-01-12
Add browser field to reduce bundle sizes

maclockard:
Set the browser field in the package.json as described here: https://github.com/defunctzombie/package-browser-field-spec to help web targeted builds find the correct version of js-yaml.

Currently when using js-yaml with webpack + prod build, js-yaml takes up 384 KB, 282 of which are due to esprima.

screen shot 2019-01-11 at 3 57 20 pm

The issue here is that webpack will use index.js as the entry point as default. Similar to #435 this will help webpack find the web version of js-yaml. The difference being that the browser field will not interfere with non web related builds. 

See https://webpack.js.org/configuration/resolve/#resolve-mainfields and https://github.com/defunctzombie/package-browser-field-spec for more info.

Here's what js-yaml looks like after setting this in my bundle:

screen shot 2019-01-11 at 4 24 53 pm

Down to 105 KB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "bin/",
     "dist/"
   ],
+  "browser": "dist/js-yaml.js",
   "bin": {
     "js-yaml": "bin/js-yaml.js"
   },


maclockard:
@puzrin whats are your thoughts on this?

puzrin:
@maclockard `dist` is legacy thing, been added when bower was alive. I'd like to remove this folder sometime, instead of more deep coupling.

maclockard:
Why not merge this for now while you work out how you want to build multiple targets for node and web? Its a pretty trivial one line change to merge/revert.

puzrin:
Because reverting will be a breaking change, and i don't like to add more maintenance problems for the future.

maclockard:
You could also in the future if the web and node build are the same, simply point browser/main to the same output to avoid the breaking change. This would be still be helpful to clearly communicate that the same build is intended to be used both for in the browser and node.

I think there is a chance that the solution you come up with to support both web and node would result in two different builds. If that ends up being the case, you would still need to use the `browser` field.

vimtaai:
Adding this would also make CDNs handle the package better. CDNs by default serve the file marked in the browser field of package.json. It would help a lot to make easier for loading dependencies from CDN based on a package's dependency field.

pelotom:
@puzrin I really don't understand your antipathy to such an innocuous change that would greatly help many users of this package. What is the "maintenance burden" here really? It's a single primitive property in the `package.json` that has only positive effects.

rbrcsk:
It's a real bummer that this isn't merged...

For me, js-yaml is a transitive dependency, and it's not a pleasent experience to track down where esprima comes from. And currently the readme is not helpful at all either:

`!!js/function in browser bundle will not work by default. If you really need it - load esprima parser first (via amd or directly).`

I had absolutely no idea what it meant, and I suspect that most other people don't either.

puzrin:
Close in favor of #592

Issue #468, opened 2019-02-18
How to control indents via param options to dump?


diberry:
An extra 2 spaces is getting added on the item "C#" and its href. How do I turn this off? 

original yaml:

`- name: LUIS Documentation
  href: index.yml
- name: Overview
  expanded: true
  items:
  - name: C#
    href: https://cognitive.uservoice.com/knowledgebase/topics/asdfasdfasdfasdfasdfasdf/127872-luis-language-understanding-intelligent-service`

Final yaml where last 2 lines have additional (but unwanted) 2 spaces:

`- name: LUIS Documentation
  href: index.yml
- name: Overview
  expanded: true
  items:
    - name: 'C#'
      href: 'https://cognitive.uservoice.com/knowledgebase/topics/asdfasdfasdfasdfasdfasdf/127872-luis-language-understanding-intelligent-service'`

puzrin:
https://github.com/nodeca/js-yaml#safedump-object---options-

`noArrayIndent` ?

diberry:
Here are all my options, I toggled noArrayIndent and the results are in the screenshot. Left = original

`
options={
            "skipInvalid":false,
            "noCompatMode":false,
            "sortKeys":false,
            "lineWidth":3000,
            "noArrayIndent":false,
            "noRefs":true,
            "flowLevel":-1,
            "condenseFlow":false
        }
`

![no-array-index-results](https://user-images.githubusercontent.com/41597107/52958567-2e10e880-3349-11e9-9a6c-54796facd9c4.PNG)

puzrin:
Is this this a question or not? Is this issue solved?

perlpunk:
@diberry could you paste your results as text instead of a screenshot? just put three backticks in one line
before and after the pasted snippet. or indent the whle snippet with 4 spaces. makes it easier to copy snippets and try them out.

@puzrin it looks like a bug to me. in the picture on the right the YAML seems invalid. items like `href: ...` in line 2 and `expanded: ...` in line 4 have no indent

diberry:
@perlpunk Thanks for the pointer.

original 

```
- name: LUIS Documentation
  href: index.yml
- name: Overview
  expanded: true
  items:
  - name: C#
    href: https://cognitive.uservoice.com/knowledgebase/topics/asdfasdfasdfasdfasdfasdf/127872-luis-language-understanding-intelligent-service
```

"noArrayIndent"= false

```
- name: LUIS Documentation
  href: index.yml
[...]

PR #469, opened 2019-02-24
 Fix for issue #468 includes passing test

diberry:

### DIFF
--- a/test/issues/0468.js
+++ b/test/issues/0468.js
@@ -8,13 +8,22 @@ test('should not indent arrays an extra level when disabled', function () {
   /* eslint-disable max-len */
   var output = yaml.dump(
     [
-      { a:'a_val', b:'b_val' },
-      { a:'a2_val', expanded:true, items:
-        [
-          { a:'a_a_val', b:'a_b_val' }
+      {
+        a: 'a_val',
+        b: 'b_val'
+      },
+      {
+        a: 'a2_val',
+        items: [
+          {
+            a: 'a_a_val',
+            b: 'a_b_val'
+          }
         ]
       }
-    ], { noArrayIndent: true });
-  var expected = '- a: a_val\n  b: b_val\n- a: a2_val\n  expanded: true\n  items:\n  - a: a_a_val\n    b: a_b_val\n';
+    ],
+    { noArrayIndent: true }
+  );
+  var expected = '- a: a_val\n  b: b_val\n- a: a2_val\n  items:\n  - a: a_a_val\n    b: a_b_val\n';
   assert.strictEqual(output, expected);
 });


test/issues/0468.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+
+var assert = require('assert');
+var yaml = require('../..');
+
+test('should not indent arrays an extra level when disabled', function () {
+  var output = yaml.dump([{"a":"a_value","b":"b_value"},{"a":"a2_value","expanded":true,"items":[{"a":"a_a_value","b":"a_b_value"}]}], { noArrayIndent: true });
```
puzrin: It's better to write such objects in nested form, starting from newline:

```js
var output = yaml.dump(
  [
    {
      a: "a_value",
....
  ],
```

Also probably worth shorten `a_value` => `a_val` to minimize useless noise.

test/issues/0468.js
```diff
@@ -0,0 +1,11 @@
...
+  var expected = '- a: a_value\n  b: b_value\n- a: a2_value\n  expanded: true\n  items:\n  - a: a_a_value\n    b: a_b_value\n';
```
puzrin: after `value` => `val`, if line length still too big, add `/* eslint-disable max-len */` above line with problem.

diberry: I will fix.

test/issues/0468.js
diberry: I will fix.

diberry:
@puzrin I ran from Makefile and saw the linter errors but wanted feedback before applying linting fixes.

.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
+{
```
puzrin: I guess, this file is not needed for this PR.

test/issues/0468.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+
+var assert = require('assert');
+var yaml = require('../..');
+
+test('should not indent arrays an extra level when disabled', function () {
+  /* eslint-disable max-len */
+  var output = yaml.dump([ { a:'a_val', b:'b_val' }, { a:'a2_val', expanded:true, items: [ { a:'a_a_val', b:'a_b_val' } ] } ], { noArrayIndent: true });
```
puzrin: Please, format this to multiline form first, as been requested (i provided sample). Lint rules can be disabled only when following those is not acceptable. Not this case.

diberry:
#sign-off

test/issues/0468.js
```diff
@@ -6,7 +6,15 @@ var yaml = require('../..');
...
-  var output = yaml.dump([ { a:'a_val', b:'b_val' }, { a:'a2_val', expanded:true, items: [ { a:'a_a_val', b:'a_b_val' } ] } ], { noArrayIndent: true });
+  var output = yaml.dump(
+    [
+      { a:'a_val', b:'b_val' },
+      { a:'a2_val', expanded:true, items:
+        [
+          { a:'a_a_val', b:'a_b_val' }
+        ]
+      }
+    ], { noArrayIndent: true });
```
puzrin:
```js
var output = yaml.dump(
  [
    {
      a: 'a_val',
      b:'b_val'
    },
    {
      a: 'a2_val',
      expanded:true,
      items: [
        {
          a: 'a_a_val',
          b: 'a_b_val'
        }
      ]
    }
  ],
  { noArrayIndent: true }
);
```

1. Change formatting to above please.
2. Is this sample really minimal required for this test? For example, i don't understand what field `expanded` does here.

diberry: fixed

diberry:
@puzrin Please let me know if there are other changes you need.

jacob-hd:
@diberry @puzrin Sorry I missed this edge case... thanks for implementing the fix.

Issue #470, opened 2019-02-25
Dumping add quote to string even it is not necessary


joshuaavalon:
As title.

### Test Code

```javascript
const jsYaml = require("js-yaml");
const fs = require("fs");
const yamlStr = jsYaml.dump({ url: "https://github.com/nodeca/js-yaml" });
fs.writeFileSync("foo.yml", yamlStr, { encoding: "utf8" });
```
### Expected Result

```yaml
url: https://github.com/nodeca/js-yaml
```

### Actual Result

```yaml
url: 'https://github.com/nodeca/js-yaml'
```

### Problem 

https://github.com/nodeca/js-yaml/blob/b6d260927b875404eac8b5e610f41d2f8fbe2425/lib/js-yaml/dumper.js#L294

https://github.com/nodeca/js-yaml/blob/b6d260927b875404eac8b5e610f41d2f8fbe2425/lib/js-yaml/dumper.js#L192-L205

All of these characters ( except`0xFEFF` ) are safe inside string. Only colon cannot be followed by space or placed at the end of line.

diberry:
This is the same as #466. Perhaps we can have a new setting to handle the quoting issue?

puzrin:
@joshuaavalon "unquoting" quality is limited only by authors time. If anyone has time to improve it - PR is welcome.

diberry:
@puzrin I'll make a PR this weekend, if no one does it before then.

diberry:
[PR 454 seems related](https://github.com/nodeca/js-yaml/pull/454)

diberry:
@puzrin Are these tests all valid/correct? 

`test.only('should not unnecessaryly apply quotes', function () {

  var expected = 'url: https://github.com/nodeca/js-yaml\n';
  var actual = yaml.dump({ url: 'https://github.com/nodeca/js-yaml' });

  assert.strictEqual(actual, expected);
});
test.only('should not unnecessaryly apply quotes - ', function () {

  var expected = 'url: https://github.com/nodeca/js-yaml\n';

  var obj = {};
  obj['url']='https://github.com/nodeca/js-yaml';

  var actual = yaml.dump(obj);

  assert.strictEqual(actual, expected);
});

test.only('should not unnecessaryly apply quotes - space then /\n at end of value', function () {

  var expected = 'url: \'https://github.com/nodeca/js-yaml \'\n';

  var obj = {};
  obj['url']='https://github.com/nodeca/js-yaml ';

[...]

Issue #123, opened 2014-05-19
Incorrect handling of a few specific regular expressions


Qantas94Heavy:
When you do something like

```
!!js/regexp /efe/g/g
```

This is an invalid JavaScript RegExp literal, however it gets parsed like the following:

```
new RegExp('/efe/g', 'g');
```

Should this not be treated literally, like in other examples such as this?

```
!!js/regexp blahblahblah
```

I think it should be treated as the following:

```
new RegExp('/efe/g/g');
```

isaacs:
I'm seeing a test failure related to this issue.

```
> yaml.load('!!js/regexp /fo/g/g')
/fo\/g/g
> process.versions
{ http_parser: '2.3.0',
  node: '1.4.2',
  v8: '4.1.0.21',
  uv: '1.4.2',
  zlib: '1.2.8',
  ares: '1.10.0-DEV',
  modules: '43',
  openssl: '1.0.1k' }
```

It seems like it is expecting `'/fo/g/g'` instead of `'/fo\/g/g'`

isaacs:
Indeed, this is working on Node 0.10, but not on 0.12 or io.js 1.x.  A change to the V8 RegExp toString method, perhaps?

puzrin:
Just installed 0.12, deleted node_modules & installed again.

`make test` is ok. Ubuntu 14.04 LTS.

isaacs:
Ah, sorry, you're right, it's only in io.js, not in node 0.12.  Almost certainly due to a V8 change.  `AssertionError: /fo\/g/g == '/fo/g/g'`.

isaacs:
```
$ nave use 0.12 node -p 'require("./").load("!!js/regexp /fo/g/g")'
/fo/g/g

$ nave use 1.5 node -p 'require("./").load("!!js/regexp /fo/g/g")'
/fo\/g/g
```

isaacs:
Confirmed:

```
$ nave use 0.12 node -p '[new RegExp("fo/g", "g").toString(), /fo\/g/g.toString()]'
[ '/fo/g/g', '/fo\\/g/g' ]

$ nave use 1.5 node -p '[new RegExp("fo/g", "g").toString(), /fo\/g/g.toString()]'
[ '/fo\\/g/g', '/fo\\/g/g' ]
```

Apparently, the `toString()` of a `new RegExp(...)` used to be different from the `toString()` method on a RegExp literal.

[...]

PR #473, opened 2019-03-10
#470 fix

diberry:


test/issues/0470.js
```diff
@@ -0,0 +1,46 @@
...
+  var expected = 'url: https://github.com/nodeca/js-yaml\n';
+  var actual = yaml.dump(
+    {
+      url: 'https://github.com/nodeca/js-yaml'
+    }
+  );
+
+  assert.strictEqual(actual, expected);
+});
+test('should not unnecessaryly apply quotes - ', function () {
```
puzrin: This test looks like previous, with no difference.

test/issues/0470.js
```diff
@@ -0,0 +1,46 @@
...
+
+  var obj = {};
+  obj['url'] = 'https://github.com/nodeca/js-yaml';
+
+  var actual = yaml.dump(obj);
+
+  assert.strictEqual(actual, expected);
+});
+
+test('should not unnecessaryly apply quotes - space then /\n at end of value', function () {
```
puzrin: Don't see difference with first test

test/issues/0470.js
```diff
@@ -0,0 +1,46 @@
...
+  var expected = 'url: \'https://github.com/nodeca/js-yaml \'\n';
+
+  var obj = {};
+  obj['url'] = 'https://github.com/nodeca/js-yaml ';
+
+  var actual = yaml.dump(obj);
+
+  assert.strictEqual(actual, expected);
+});
+test('should not unnecessaryly apply quotes - space after colon', function () {
```
puzrin: Don't see difference with first test

test/issues/0470.js
```diff
@@ -0,0 +1,46 @@
...
+test('should not unnecessaryly apply quotes', function () {
+
+  var expected = 'url: https://github.com/nodeca/js-yaml\n';
+  var actual = yaml.dump(
+    {
+      url: 'https://github.com/nodeca/js-yaml'
+    }
+  );
+
+  assert.strictEqual(actual, expected);
```
puzrin: That's not enougth. Need to check that after load back result will be exact as initial

lib/js-yaml/dumper.js
```diff
@@ -291,12 +284,14 @@ function chooseScalarStyle(string, singleLineOnly, indentPerLevel, lineWidth, te
       } else if (!isPrintable(char)) {
         return STYLE_DOUBLE;
       }
-      plain = plain && isPlainSafe(char);
+      if (!isPlainSafe(char)) {
+        plain = plain && !isWhitespace(string.charCodeAt(i + 1));
+      }
```
puzrin: This is a very cryptic way to express "if plainsafe and next is not whitespace". Very deficult to understand real intent. Please, reorganize somehow.

lib/js-yaml/dumper.js
```diff
@@ -272,7 +265,7 @@ function chooseScalarStyle(string, singleLineOnly, indentPerLevel, lineWidth, te
       if (!isPrintable(char)) {
         return STYLE_DOUBLE;
       }
-      plain = plain && isPlainSafe(char);
+      plain = plain && (isPlainSafe(char) && !isWhitespace(string.charCodeAt(i + 1)));
```
puzrin: `isWhitespace(string.charCodeAt(i + 1))` missed length check, to avoid out of bounds access at [i + 1]. Such thing will trash JIT optimisations.

lib/js-yaml/dumper.js
```diff
@@ -194,14 +194,7 @@ function isPlainSafe(c) {
...
     // - c-flow-indicator
-    && c !== CHAR_COMMA
-    && c !== CHAR_LEFT_SQUARE_BRACKET
-    && c !== CHAR_RIGHT_SQUARE_BRACKET
-    && c !== CHAR_LEFT_CURLY_BRACKET
-    && c !== CHAR_RIGHT_CURLY_BRACKET
-    // - ":" - "#"
-    && c !== CHAR_COLON
-    && c !== CHAR_SHARP;
+    && c !== CHAR_COLON;
```
puzrin: It would be nice to cover this change with tests. For example, check dumping (and load back) of this strings:

`C#`, `C#123`, `C #123`, `C# 123`

lib/js-yaml/dumper.js
puzrin: Should it check next or prev char?

`{ test: 'C #123' }` - make sure this will be quoted.

GerHobbelt:
https://github.com/diberry/js-yaml/pull/1 should answer many of the comments above. That pullreq introduces round-trip tests and has more tests.

puzrin:
@GerHobbelt i think it's better to make your PR as independent, and to this repo. Then we can discuss details.

puzrin:
Close as incomplete/invalid.

As been noted in https://github.com/nodeca/js-yaml/pull/484#issuecomment-483705876, this will break flow style of dump.

PR #480, opened 2019-04-05
Fix possible code execution in (already unsafe) load()

rlidwka:
Object with executable toString() property when used as a map key will execute that function.

This happens only for `load()`, which should not be used with untrusted data anyway. `safeLoad()` is not affected because it can't parse functions.

Example:

```yaml
$ cat test.yaml
{ toString: !<tag:yaml.org,2002:js/function> 'function (){return Date.now()}' } : 1

$ ./bin/js-yaml.js ./test.yaml 
{
  "1553107949161": 1
}
```

After this PR it would return `[object Object]` as if toString wasn't there.

Dates are not affected, and if you for whatever reason are using custom types as keys (which is **not** supported very well), you can change internal class with `Symbol.toStringTag` bypassing added validation check.


ddillard:
Is anyone getting a CVE for this issue?  If not, I can do it.

viccastro0916 (commented):
Address this

PR #492, opened 2019-05-01
feat: allow type to be dumped as a different type

jaulz:
With this change it will be possible to dump a type to a different type based on the chosen style. The `predicate` function can decided based on it's style whether it can represent it or not.

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -658,16 +658,15 @@ function detectType(state, object, explicit) {
 
   for (index = 0, length = typeList.length; index < length; index += 1) {
     type = typeList[index];
+    style = state.styleMap[type.tag] || type.defaultStyle;
 
     if ((type.instanceOf  || type.predicate) &&
         (!type.instanceOf || ((typeof object === 'object') && (object instanceof type.instanceOf))) &&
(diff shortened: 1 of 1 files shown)


puzrin:
This feature looks strange. Could you explain why it may be useful for all to hack types on the fly?

jaulz:
@puzrin for example "!square 4" could have two styles: "original" and "calculated". If you dump it with style "original" it will dump as "!square 4" again but if you dump it with style "calculated" it should dump as "16" only.

puzrin:
As far as i understand, loader should be able to load back dumped result. If you dump `16`, that will be a number after load, not custom type. Seems you try to use yaml as "programming language", that's attractive, but not good idea.

jaulz:
@puzrin actually, I would like to use it as a preprocessor to generate a final yaml file that only needs built in types and no custom types. Any suggestion for any other option?

puzrin:
IMHO it's better to keep things separate and prepare data to final state before dump.

jaulz:
Mh, okay, so there's no chance to merge this? If yes, do you know other tools I could use for preprocessing?

puzrin:
I tend to reject as "out of project scope".

jaulz:
Too bad... thanks anyway for the input!

Issue #414, opened 2018-04-09
Getting 'Unexpected identifier' when function has await 


kiran-z:
Unable to load the .yml when a function has await keyword.

``` 
function: !!js/function > 
            async function() {
              let b = new Promise(function(resolve, reject) {
                resolve('a');
              })
              await b;
          }
```

`SyntaxError: Unexpected identifier
    at new Function (<anonymous>)
    at Type.constructJavascriptFunction [as construct] (/Users/admin/node_modules/js-yaml/lib/js-yaml/type/js/function.js:69:10)`

Are AsyncFunctions supported?

puzrin:
1. You have wrong location of last comma. Correct will be [this](http://nodeca.github.io/js-yaml/#yaml=ZnVuY3Rpb246ICEhanMvZnVuY3Rpb24gPiAKICAgICAgICAgICAgYXN5bmMgZnVuY3Rpb24oKSB7CiAgICAgICAgICAgICAgbGV0IGIgPSBuZXcgUHJvbWlzZShmdW5jdGlvbihyZXNvbHZlLCByZWplY3QpIHsKICAgICAgICAgICAgICAgIHJlc29sdmUoJ2EnKTsKICAgICAgICAgICAgICB9KQogICAgICAgICAgICAgIGF3YWl0IGI7CiAgICAgICAgICAgIH0=)
2. Anyway, async/await seems not supported. PR welcome.

puzrin:
Close as rejeced: https://github.com/nodeca/js-yaml/pull/496 after talk in PR - please do that via custom schema. No wish to fill core with js-specific progamming things.

PR #496, opened 2019-05-10
Added support for async functions

GeekyDeaks:
Hi,

This is my attempt to provide support for async functions.  I encountered the same problem as #414 and figured I'd take a quick look to see if it was an easy fix.

Unfortunately, it required a change to the eslint config, which I am not entirely happy about.   Maybe it would be better to provide override statements?

### DIFF
--- a/test/12-loader-es2017.js
+++ b/test/12-loader-es2017.js
@@ -12,9 +12,6 @@ var TEST_SCHEMA = require('./support/schema').TEST_SCHEMA;
 
 suite('Loader ES2017', function () {
 
-  // skip test if we don't support ES2017
-  if (semver.lt(process.version, '8.0.0')) return;
-
   var samplesDir = path.resolve(__dirname, 'samples-es2017');
 
   fs.readdirSync(samplesDir).forEach(function (jsFile) {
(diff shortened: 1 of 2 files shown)


GeekyDeaks:
Yep - failing on older versions!

Ok, I'm going to have a think about how best to fix it, but if you have any suggestions on a suitable approach I am all ears

puzrin:
I don't like idea to move into lang-specific direction in core. IMO it's better to keep such specific thing as "plugin". Implementation allows to extend schemas as you wish.

GeekyDeaks:
I think I understand what you mean.  So we would have something like:

`function: !!js/asyncfunction 'async () => { }'`

Correct?

puzrin:
That's one on possibilities. Or you can reassemble schema, using your own implementation for `!!js/function` instead of built-in one.

GeekyDeaks:
I like the idea of the plugin too, so I'll close this PR and write one.

Would it be of any use for me to create another PR to place the plugin in `/examples` and link it to #414?

GeekyDeaks:
Sorry @puzrin, can I re-open this please?

I'm having trouble `dump()`ing using a plugin.  There seems to be an explicit check in the core code for a function:

https://github.com/nodeca/js-yaml/blob/665aadda42349dcae869f12040d9b10ef18d12da/lib/js-yaml/dumper.js#L671-L672

I'm trying to figure out how to override this in a plugin...

GeekyDeaks:
I forgot to mention - the `dump()` works correctly with the PR, so there is something going on I cannot quite fathom

GeekyDeaks:
Sorry @puzrin - it was a mistake in my plugin.  Apologies for hassling you with this.  I'll close this one now.

PR #500, opened 2019-05-21
Add Docker support

xtellurian:
This PR adds a Dockerfile and a reference to a built Docker image in Docker Hub.

Feel free to move the docker image to another user.


puzrin:
Docker image is out of project scope.

PR #517, opened 2019-09-13
feat: allow passing reference names for dupe objects when dumping

njoyard:
This PR adds a new feature to allow setting reference names when dumping:

```js
let knownObject = { foo: "bar" };
let unknownObject = { foo: "baz" };
yaml.dump(
  { first: knownObject, second: knownObject, third: unknownObject, fourth: unknownObject },
  { refMap: { "myReference": knownObject } }
);
/*
first: &myReference
  foo: bar
second: *myReference
third: &ref_1
  foo: baz
fourth: *ref_1
*/
```

I'm not sure if that would get enough interest to be merged? And whether there should be tests (I couldn't find any existing tests for refs)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,10 @@ options:
   function, use the function to sort the keys.
 - `lineWidth` _(default: `80`)_ - set max line width.
 - `noRefs` _(default: `false`)_ - if `true`, don't convert duplicate objects into references
+- `refMaps`_(default: `{}`)_ - mapping of reference names to use for duplicate
+  objects, eg. `{ myReference: myObject, ... }`.  Duplicate objects not found in
+  the mapping will have a default reference name assigned (eg. `ref_42`).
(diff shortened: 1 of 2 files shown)


rlidwka:
If we want to allow user to override reference names, callback will be more useful than map imho.

So something like this:

```js
let knownObject = { foo: "bar" };
let unknownObject = { foo: "baz" };
let cnt = 0;
yaml.dump(
  { first: knownObject, second: knownObject, third: unknownObject, fourth: unknownObject },
  { onDuplicateRef: obj => obj === knownObject ? 'myReference' : `ref_${++cnt}` }
);
```

The reason is: if you require a map from user, they'd have to iterate over object first to make it.

rlidwka:
Closing this one, because suggested API for this isn't ideal.

Not sure whether this feature is useful at all, but if it is, it should be a callback of some kind.

PR #519, opened 2019-10-11
Add equals sign to list of unsafe values for plain styling

murtazajafferji:
According to this post and my use case, the strings with equals sign needs to be quoted:
https://stackoverflow.com/questions/19109912/do-i-need-quotes-for-strings-in-yaml
Use quotes if your value includes special characters, (e.g. :, {, }, [, ], ,, &, *, #, ?, |, -, <, >, =, !, %, @, \).

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -221,7 +221,7 @@ function isPlainSafeFirst(c) {
     && c !== CHAR_RIGHT_SQUARE_BRACKET
     && c !== CHAR_LEFT_CURLY_BRACKET
     && c !== CHAR_RIGHT_CURLY_BRACKET
-    // | “#” | “&” | “*” | “!” | “|” | “=” | “>” | “'” | “"” 
+    // | “#” | “&” | “*” | “!” | “|” | “=” | “>” | “'” | “"”
     && c !== CHAR_SHARP
     && c !== CHAR_AMPERSAND
     && c !== CHAR_ASTERISK


puzrin:
Do you have a concrete sample with broken result? All such PRs should go with tests.

puzrin:
This PR will be declined if no example of which error does it fix.

murtazajafferji:
``>>> from  ruamel.yaml import YAML
>>> yaml = YAML()
>>> yaml.load("=")
<ruamel.yaml.comments.TaggedScalar object at 0x7ff7d27d9a50>
>>> yaml.load("'='")
'='

puzrin:
That example is invalid. First, this must be executable code, not something, from your browser console. Second, this code works as expected:

```
node -e 'console.log(require("./").load("="))'
=
```

in js-yaml folder.

puzrin:
If problem is with ruamel.yaml library, fix should be sent there.

murtazajafferji:
The example above and the one below is showing that the output of a safeDump from js-yaml will not be interpreted as valid yaml by the most popular yaml parsers for Python. This is for PyYAML:
```console
>>> yaml.dump(yaml.load("="))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/yaml/__init__.py", line 114, in load
    return loader.get_single_data()
  File "/usr/local/lib/python2.7/dist-packages/yaml/constructor.py", line 45, in get_single_data
    return self.construct_document(node)
  File "/usr/local/lib/python2.7/dist-packages/yaml/constructor.py", line 49, in construct_document
    data = self.construct_object(node)
  File "/usr/local/lib/python2.7/dist-packages/yaml/constructor.py", line 94, in construct_object
    data = constructor(self, node)
  File "/usr/local/lib/python2.7/dist-packages/yaml/constructor.py", line 420, in construct_undefined
    node.start_mark)
yaml.constructor.ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<string>", line 1, column 1:
    =
    ^
>>> yaml.dump(yaml.load("'='"))
"'='\n"
```

test/issues/0519.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+var assert = require('assert');
+var yaml   = require('../../');
+
+test('Loader should not strip quotes for lines containing equals sign', function () {
```
puzrin: I'm a bit confused about test name. If problem is in dumper (missed quotes for `=`), why title is bout testing loader?

lib/js-yaml/dumper.js
```diff
@@ -220,16 +221,17 @@ function isPlainSafeFirst(c) {
...
+    // | “#” | “&” | “*” | “!” | “|” | “=” | “>” | “'”
     && c !== CHAR_SHARP
     && c !== CHAR_AMPERSAND
     && c !== CHAR_ASTERISK
     && c !== CHAR_EXCLAMATION
     && c !== CHAR_VERTICAL_LINE
+    && c !== CHAR_EQUALS
     && c !== CHAR_GREATER_THAN
     && c !== CHAR_SINGLE_QUOTE
+    // | “"” | “%” | “@” | “`”)
```
puzrin: Why comment & `"` location was changed?

test/issues/0519.js
```diff
@@ -0,0 +1,10 @@
...
+  var line_with_equals_sign = "'='\n";
+  var result_of_load_and_dump = yaml.dump(yaml.load(line_with_equals_sign));
+  assert.strictEqual(line_with_equals_sign, result_of_load_and_dump);
+});
```
puzrin: IMO actual test content could be:

```js
// pyyaml fails with unquoted `=`
// https://yaml-online-parser.appspot.com/?yaml=%3D%0A&type=json
assert.strictEqual(yaml.load(yaml.dump('=')), '=');
assert.strictEqual(yaml.dump('='), "'='");
```

lib/js-yaml/dumper.js
murtazajafferji: I moved it so that the previous comment didn't get too long for readability. I can undo this change if you think it's unnecessary.

test/issues/0519.js
murtazajafferji: Thanks for catching that! I update the name to be more clear.

test/issues/0519.js
murtazajafferji: Thanks for the suggestion! I updated the test.

lib/js-yaml/dumper.js
puzrin: Please, roll back this change, to reduce confusion about PR intent.

murtazajafferji:
@puzrin I have addressed all the comments. Do you have any other suggestions or can we complete this PR?

puzrin:
Hm, i don't know why my last note is invisible. Could you return those comments and locations back?

- https://github.com/nodeca/js-yaml/pull/519/files#diff-fff1647e8e19b34e8c0a153680cd8723R224
- https://github.com/nodeca/js-yaml/pull/519/files#diff-fff1647e8e19b34e8c0a153680cd8723R233

The only change should be added `=` to first one.

lib/js-yaml/dumper.js
murtazajafferji: Done!

murtazajafferji:
@puzrin Done! Btw, your comment was collapsed by default because the comment thread was marked as resolved. If you wanted it to be shown, you could press "Unresolve conversation."

PR #521, opened 2019-10-21
dumper: don't quote strings with # without need

vcache:
According to [spec](https://yaml.org/spec/1.2/spec.html#id2788859) a sharp symbol may be used  within a key with plain style (i.e. without quotes) in certain conditions:
```
[130]   ns-plain-char(c) ::=   ( ns-plain-safe(c) - “:” - “#” )
                             | ( /* An ns-char preceding */ “#” )
                             | ( “:” /* Followed by an ns-plain-safe(c) */ )
```
That's handy when key contains URI like so:
```
uris:
  - http://example.com/page#anchor: some-example
```

### DIFF
--- a/lib/js-yaml/dumper.js
+++ b/lib/js-yaml/dumper.js
@@ -12,6 +12,7 @@ var _hasOwnProperty = Object.prototype.hasOwnProperty;
 
 var CHAR_TAB                  = 0x09; /* Tab */
 var CHAR_LINE_FEED            = 0x0A; /* LF */
+var CHAR_CARRIAGE_RETURN      = 0x0D; /* CR */
 var CHAR_SPACE                = 0x20; /* Space */
 var CHAR_EXCLAMATION          = 0x21; /* ! */
 var CHAR_DOUBLE_QUOTE         = 0x22; /* " */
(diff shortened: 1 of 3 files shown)


puzrin:
1. Term "fix" usually means something was broken (invalid output). As far as i understand, output without this update is valid. So, that's not fix but improvement.
2. Need test to demonstrate use cases, why this change done.

lib/js-yaml/dumper.js
```diff
@@ -201,8 +201,9 @@ function isPlainSafe(c) {
     && c !== CHAR_LEFT_CURLY_BRACKET
     && c !== CHAR_RIGHT_CURLY_BRACKET
     // - ":" - "#"
+    // /* An ns-char preceding */ "#"
     && c !== CHAR_COLON
-    && c !== CHAR_SHARP;
+    && (c !== CHAR_SHARP || prev && isPrintable(prev));
 }
```
puzrin: `((c !== CHAR_SHARP) || (prev && isPrintable(prev)));`

I'd use more scopes to simplify understanding.

vcache:
Thanks for your review. You right, actually it is not a bug fix but proposition for more efficient usage of YAML specification. I've added some tests and made code more readable.

puzrin:
- please move tests [here](https://github.com/nodeca/js-yaml/tree/master/test/issues) with Issue/PR number
- squash all and change commit name to "dumper: don' quote strings with # without need" - that's more informative.

Everything else seems ok.

puzrin:
Thank you!

Himura2la:
The test does not work. You should not parse the sample YAML if you are testing how it's dumped. I changed the test like this:

```js
test('Don\'t quote strings with # without need', function () {
  var required = readFileSync(require('path').join(__dirname, '/0521.yml'), 'utf8');
  var data = {
    'http://example.com/page#anchor': 'no#quotes#required',
    'parameter#fallback': 'quotes #required',
    'foo #bar': 'key is quoted'
  };
  var actual = yaml.safeDump(data);
  assert.equal(actual, required);
});
```

and it fails. because the `http://example.com/page#anchor` key is being quoted (because of a colon)

PR #522, opened 2019-10-25
Documentation for the listener option and a small improvement for it.

ZPiDER666:
I needed some mini-AST-features for a project and after adding these features i stumbled upon the listener option, so i documented it. To make the integration more straight-forward i also added a "startline" parameter to the "close" listener callback.


puzrin:
Listener option is hidden, because we don't like to promote it. It's sertaintly NOT for readme. Probably, it could be added in examples, but with appropriate comment inside.

PR #523, opened 2019-11-01
Add unpkg and jsdelivr fields to point to browser build

tmcw:
This adds unpkg & jsdelivr fields to package.json so that unpkg & jsdelivr can automatically find the right build to serve for browsers (and so can tools like d3-require).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,8 @@
   "bin": {
     "js-yaml": "bin/js-yaml.js"
   },
+  "unpkg": "dist/js-yaml.min.js",
+  "jsdelivr": "dist/js-yaml.min.js",
   "dependencies": {
     "argparse": "^1.0.7",
     "esprima": "^4.0.0"


puzrin:
Hm... seems this is not actual:

- https://www.jsdelivr.com/package/npm/js-yaml jsedlivr serves all conent of npm
- unpkg has no doc with this option.

Could you comment? I plan to cleanup garbage in 4.0

tmcw:
Sure, if you look at the page https://www.jsdelivr.com/package/npm/js-yaml, the `dist` file is at the top and is marked as the "default file", and if you follow the patterns from the [jsdelivr homepage](https://www.jsdelivr.com/) and use the "default file" form, you get the correct file: https://cdn.jsdelivr.net/npm/js-yaml - and in versions before that PR, you don't: https://cdn.jsdelivr.net/npm/js-yaml@2

Similarly, https://unpkg.com/js-yaml refers to the correct default file and in previous versions, it refers incorrectly to the commonjs entry point: https://unpkg.com/js-yaml@2

Both unpkg & js-yaml have listing pages like the one you link, but the benefit of this PR is that their "default file" endpoints - what d3-require uses, and what people can use in their script tags - need this additional information to refer to the correct dist.

puzrin:
Ah, i'm not familiar with "serverless" webdev. Always thinked people use something like `https://cdn.jsdelivr.net/npm/js-yaml@3/dist/js-yaml.min.js` in script tag or web loaders.

What happens when i add `.mjs` build? `/dist` usually contains multiple builds like this: https://github.com/nodeca/pako/tree/master/dist. Does a partial [incomplete] solution worth keeping such indirect magic?

tmcw:
Yes - people should just use that URL in their script tag, to avoid the redirect. And the jsdelivr field in package.json helps them do that - before this PR, the page for js-yaml ([here's the page for 3.1.0](https://www.jsdelivr.com/package/npm/js-yaml?version=3.1.0)) would incorrectly recommend that they use the commonjs file in their script tag.

Mjs builds tend to be opt-in with all the systems and they all behave differently. unpkg's ?module option will use the `module` field in package.json instead of the unkpg field, and it looks like jsdelivr splits that off into a separate service, esm.run, which will also use the other field https://www.jsdelivr.com/esm

Issue #526, opened 2019-11-14
Wrong quote position writing condensed flow


GabrieleC:
This document:
```
key1: val1
key2: val2
```
dumped with options `{ flowLevel: 0, condenseFlow: true }` produces output with misplaced quote:
```
{"key1":val1", key2":val2}
```

puzrin:
Dupe of #412

GabrieleC:
I made a simple bugfix for this, should i discard it or make a PR?

puzrin:
I guess those options are mutually exclusive. `condenseFlow` was added for concrete case - URL encoding, and i would not like to spread it wide without need. If you wish this to be changed, this should be discussed first.

GabrieleC:
When using `condenseFlow: true`:
 - if `flowLevel == -1` (default value) all collections will be write as blocks, so `condenseFlow: true` will be useless.
 - if `flowLevel >= 0` there will be the bug.

So either "condenseFlow" is useless or it's broken.

There's something I'm misunderstanding?

puzrin:
We do nod add features "just for fun". As i said, this option was introduced for concrete use case, and not expected to be used anyhow else. If you wish this to be changed - describe your use case first, before any coding.

GabrieleC:
The use case I'm referring is URL encoding. `condenseFlow` is intended to be used with `flowLevel = 0`, otherwise it would be useless. An example is shown in the original issue: https://github.com/nodeca/js-yaml/issues/346.

The bug is in dumper.js, line 572, where comma is appended after the quote instead of before.

I understand this library isn't intended to be used as minifier, so if this fix isn't well accepted I will not go further.

puzrin:
Probably i missunderstood you. Take a look at https://github.com/nodeca/js-yaml/pull/358/files, there are some tests. If something is missed and works not as intended in original PR, fix will be accepted.

Just make sure that's really a bug, because this option should be actively used by PR authors and they did not reported problem before.

GabrieleC:
The original test doesn't throw error because it uses a collection with only one item, so no comma is written. 

I think this bug wasn't included in the original PR (#358), because code is much different from the current version. Maybe PR authors never upgraded to a newer version with the bug.

I made a PR including the fix and an improvement to the original test to intercept the bug:  https://github.com/nodeca/js-yaml/pull/358/files

PR #527, opened 2019-11-14
Fix issue #526: wrong quote position writing condensed flow

GabrieleC:


lib/js-yaml/dumper.js
```diff
@@ -569,9 +569,8 @@ function writeFlowMapping(state, level, object) {
       pairBuffer;
 
   for (index = 0, length = objectKeyList.length; index < length; index += 1) {
-    pairBuffer = state.condenseFlow ? '"' : '';
-
-    if (index !== 0) pairBuffer += ', ';
+    pairBuffer = index === 0 ? '' : ', ';
```
puzrin: Please avoid such notation, it's not enough easy to read. Use `if (index !== 0) pairBuffer += ', ';` and so on in appropriate places.

puzrin:
Seems ok in general. Please take a look at comment & squash all to single commit.

GabrieleC:
> 
> 
> Seems ok in general. Please take a look at comment & squash all to single commit.

@puzrin Improved readability and squashed commits.

puzrin:
Thank you very much!

PR #528, opened 2019-11-16
Adds style option 'nestedSeqBlocks' for nested sequence

AlexHolly:
Each nested array will be in a new line if true
```
-
  - 1
```

else
`- - 1`


puzrin:
Solved in master (upcoming v5). This style is supported

PR #535, opened 2020-02-17
Added missing value warning

AlexHolly:
```jsYaml.safeLoad('test:', { 'onWarning': (err) => { throw err  } })```

Input
`test:`

output
```
Missing value
at line 1
```

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -10,6 +10,7 @@ var DEFAULT_FULL_SCHEMA = require('./schema/default_full');
 
 
 var _hasOwnProperty = Object.prototype.hasOwnProperty;
+var showHintOnMissingValue = false;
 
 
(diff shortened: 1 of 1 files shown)


puzrin:
IMO it's better to leave this to linter tool. There is nothing wrong with this notation.

AlexHolly:
Maybe an additional event like 'onHint'? I don't want to use a linter only for this special case.

puzrin:
> Maybe an additional event like 'onHint'? I don't want to use a linter only for this special case.

That's not enough for mainstream changes.  `onXXX` are kludges, and i strongly against adding new ones.

- "I need that" - bad agument.
- "Many users need that" - better, at least.

At first glance, suggested change is personal opinion based, with alternate solution (linter) already available. If i missed something - let me know.

PR #537, opened 2020-02-29
mult yaml document parse support

everhopingandwaiting:
mult yaml document parse support

### DIFF
--- a/index.html
+++ b/index.html
@@ -28,7 +28,7 @@
   <body>
 
     <div class="header">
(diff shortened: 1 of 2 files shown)


everhopingandwaiting:
image

puzrin:
1. I dont like idea to maintain.
   - multiple demos
   - with different technologies.
2. I'm not sure demo needs this mode at all.

This mode can be implemented via checkbox in existing demo. But anyway, need of this feature should be discussed first.

PR #547, opened 2020-05-15
fix the use of parameters

zjffun:
`safeLoad()`, `safeLoadAll()` will lose options when pass 2 parameters.

This fix is a breaking change.


ExE-Boss:
<https://github.com/nodeca/js-yaml/pull/548> adds support for this without breaking backwards compatibility.

zjffun:
Your fix is better.

PR #548, opened 2020-05-16
fix(loader): Add support for `safe/loadAll(input, options)`

ExE-Boss:
This makes it possible to invoke `safeLoadAll` and `loadAll` by passing only `input` and `options` as two arguments without breaking support for the 3‑arguements invocation with a nullish `iterator` parameter.

---

Supersedes and closes <https://github.com/nodeca/js-yaml/pull/381>
Supersedes and closes <https://github.com/nodeca/js-yaml/pull/547>


rlidwka:
Merged, thanks!

PR #557, opened 2020-06-15
Improve Plain Style support

Himura2la:
The test in #521 is invalid, and did not test the real quotation cases.

I fixed the test and changed the `isPlainSafe` function so that it follows the [specs](https://yaml.org/spec/1.2/spec.html#ns-plain-safe-out) more precisely.

Although, I still don't understand why specs exclude  **s-white** from **ns-plain-char**:

```
[34]            ns-char ::= nb-char - s-white
[127]  ns-plain-safe(c) ::= c = flow-out ⇒ ns-plain-safe-out
                            ...
[128] ns-plain-safe-out ::= ns-char
[130]  ns-plain-char(c) ::= ( ns-plain-safe(c) - “:” - “#” )
                            ...

ns-plain-char(flow-out) ::= nb-char - s-white - “:” - “#”
```

But we don't see it even in examples:

```
- Up, up, and away!
```

 I included whitespaces into Plain scalars as they were before my changes


rlidwka:
Merged into `dev` branch as https://github.com/nodeca/js-yaml/commit/91f9662250a55ec30ffa2948da6c331efbe2364d (after resolving conflicts), gonna be released as `4.0` later. Thanks!

PR #558, opened 2020-06-20
feat: add mjs support

xiaoxiangmoe:
Allow users to use these in mjs

```js
import { safeLoad } from 'js-yaml';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "repository": "nodeca/js-yaml",
   "files": [
     "index.js",
+    "index.mjs",
     "lib/",
     "bin/",
     "dist/"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,6 +29,13 @@
   },
   "unpkg": "dist/js-yaml.min.js",
   "jsdelivr": "dist/js-yaml.min.js",
+  "exports": {
+    ".": {
+      "import": "./index.mjs",
+      "require": "./index.js",
+      "default": "./index.js"
+    }
+  },
```
realityking, line 38: I suggest to also export `package.json`. Otherwise some tools might trip: https://github.com/nodejs/node/issues/33460

puzrin:
Fixed by https://github.com/nodeca/js-yaml/commit/1b5fea8526fea734041334db3687ae5a919a0dfc

PR #561, opened 2020-06-30
support reading of properties on implicit block mapping keys

dota17:
now , the following yaml document can be loaded
```yaml
&anchor foo:
  foo: bar
  *anchor: duplicate key
  baz: bat
  *anchor: duplicate key
```


perlpunk:
Note that the example is invalid YAML.
There needs to be a space after an alias key before the colon (because a colon can be part of an anchor/alias).
It should be:
```
  *anchor : duplicate key
```

rlidwka:
`pyyaml` parses this, however http://ben-kiki.org/ypaste/ (which is specified as reference parser) does not.

Gonna check it again later, but I think we should require a space after colon there?

perlpunk:
pyyaml and libyaml can parse it, but that it actually a bug.

yuriy-yarosh:
And what about object / array anchor merging ?
Is it a valid syntax for Yaml 1.2 ?

```yaml
"@figen/config-eslint@*":
  dependencies: &eslint
    "eslint": "8.55.0"
    
"@figen/config-typescript@*":
  dependencies: &typescript
    "typescript": "^5.3.3"
      
"figen@*":
  dependencies:
    << : {*eslint, *typescript}
```

perlpunk:
Correct would be:
```
<< : [*eslint, *typescript]
```
The syntax itself is correct. The parser doesn't know anything about `<<` - it is just a string at this point.
The constructor detects it as a merge key, and that depends on library if it allows this for YAML 1.2. It should not be enabled by default.

puzrin:
https://github.com/nodeca/js-yaml/commit/a31026594c28989144161596ce78a16f810ec6e5 fixed in more robust way (properly keep anchorMap content on roll-backs)

PR #568, opened 2020-07-22
add range feature

rdeline:
For some tools, it's useful to know the YAML source range (start index, end index) corresponding to each node in the deserialized YAML. This information is currently thrown away during parsing. This PR adds an option called `rangeProp` that lets the user specify the name of a property where the range should be stored for mappings and lists. For example, if the user invokes the parser like this
```
var txt = "" +
  "food:\n" +
  "  - apple:\n" +
  "      calories: 95\n" +
  "      price: $0.40\n" +
  "  - banana:\n" +
  "      calories: 100\n" +
  "      price: $80\n"

const obj = yaml.load(txt, {rangeProp: '_range'});
```
then there will be the following additional properties added to the tree:
```
obj._range // { start: 0, end: 104 }
obj.food._range // { start: 8, end: 104 }
obj.food[0]._range // { start: 10, end: 57 }
obj.food[0].apple._range // { start: 23, end: 57 }
```
and so forth. The name of the property is completely under the caller's control and is passed as the `rangeProp` option.

This feature is particularly useful for error-reporting for tools that use YAML for configuration. With this feature, the tool can report exactly where in the config file the error/warning/info occurs.

If the user does not pass the `rangeProp` option, then the range information is thrown away, as usual.

### DIFF
--- a/lib/js-yaml/loader.js
+++ b/lib/js-yaml/loader.js
@@ -138,7 +138,6 @@ function State(input, options) {
   this.legacy    = options['legacy']    || false;
   this.json      = options['json']      || false;
   this.listener  = options['listener']  || null;
-  this.rangeProp = options['rangeProp'] || null;
 
   this.implicitTypes = this.schema.compiledImplicit;
   this.typeMap       = this.schema.compiledTypeMap;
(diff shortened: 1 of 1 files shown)


puzrin:
IMO that looks clever, but too hacky. Source mapping belongs to AST absraction layer (not existing in this parser), not to end data. I don't like idea of kluges for data flow (in scope of long term maintenance).

May be you could use hidden `listener` option? It was specially added for AST-related hacks.

rdeline:
@puzrin Good suggestion! I just updated the PR so that the range info instead generates a listener event
```
    state.listener('range', state.result, { start: start, end: state.position });
```
Then to support the use case I care about, I just write my listener like this:
```
const options = {
  listener: (etype, obj, range) => {
    if (etype === 'range') {
      obj['$range'] = range;
    }
  }
};
const obj = yaml.load(txt, options);
```
Now there is no pollution of the end data, unless that's what the caller decides to do.

rdeline:
Never mind. I now see that the existing 'open' and 'close' events are enough to get the info that I want. [face palm] Sorry!

Issue #477, opened 2019-03-23
2019-03-22 20:18:42 +0100 interpreted as a string


DirtyF:
`2019-03-13 20:18:42 +0000` is a date format considered valid in Jekyll, when switching to 11ty, it's not. (see: https://github.com/11ty/eleventy/issues/413#issuecomment-475832065)

`space separated:  2001-12-14 21:59:43.10 +0000` returns  a string on https://nodeca.github.io/js-yaml/

http://www.yamllint.com/ will add a colon after parsing: `2019-03-13 20:18:42 +00:00`

puzrin:
This package is about yml standard, not about jekyll.  If you think js-yaml does not conform spec, please give URL to spec details. Also, reference implementation is pyyaml, if you need to compare.

josephholsten:
Just found this and I've got to say, the spec is way to ambiguous on its definition of timestamps. [YAML 1.2 §10.1. Failsafe Schema](https://yaml.org/spec/1.2.2/#101-failsafe-schema) doesn't actually provide any guidance about what to do with [Example 2.22 Timestamps](https://yaml.org/spec/1.2.2/#example-timestamps) even though it's implied that it shouldn't need an explicit tag. As an editor of other specs, I'd call this a spec bug, not just an implementation one. So congrats, you're both right!

I'm inclined to say we should defer to [RFC 3339 Date and Time on the Internet: Timestamps §5.6 Internet Date/Time Format](https://www.rfc-editor.org/rfc/rfc3339#section-5.6), which covers the `canonical`, `iso8601` and `date` examples, but I really don't know what to do about `spaced: 2001-12-14 21:59:43.10 -5`. If it were me, I'd say allowing space would go from RFC 3339's:
```
   time-numoffset  = ("+" / "-") time-hour ":" time-minute
   time-offset     = "Z" / time-numoffset

   partial-time    = time-hour ":" time-minute ":" time-second
                     [time-secfrac]
   full-date       = date-fullyear "-" date-month "-" date-mday
   full-time       = partial-time time-offset

   date-time       = full-date "T" full-time
```

to instead have:
```
   full-time       = partial-time [" "] time-offset
   date-time       = full-date ("T" / "t" / " ") full-time
```

But then we don't either allow `2019-03-13 20:18:42 +0000` (no colon) or `2001-12-14 21:59:43.10 -5` (not even remotely close to a `time-offset`).

So I guess I'm off to speak with the editors!

josephholsten:
[...]

PR #579, opened 2020-09-20
Add support for ISO-8601 time offsets without colons.

mekka:
Adds support for a common ISO-8601 time zone offset format.

ISO-8601 dates can include a time offset from Zulu time, wither with a semi colon, such as "+05:00" or without a semi colon like "+0500." This pull request adds support for the format without semi-colon, as well as updates a test to confirm the new functionality.

- **With semi-colon**: 2001-12-14 21:59:43.10 -05:00
- **Without semi-colon**: 2001-12-14 21:59:43.10 -0500

Bonus: This should resolve issue #477 reported by @DirtyF. The issue isn't actually a Jekyll Issue, but is caused by js-yaml having slightly less support for ISO 8601 dates than other common YAML parsers. That being said, this is a tiny change that will make it easier for folks looking to migrate from [Jekyll](https://jekyllrb.com/) sites (including [Octopress](http://octopress.org/)), to JS based static site generators like [Eleventy](https://www.11ty.dev/). So fixing this could result in greater adoption of the js-yaml project.

### DIFF
--- a/lib/js-yaml/type/timestamp.js
+++ b/lib/js-yaml/type/timestamp.js
@@ -17,7 +17,7 @@ var YAML_TIMESTAMP_REGEXP = new RegExp(
   ':([0-9][0-9])'                    + // [6] second
   '(?:\\.([0-9]*))?'                 + // [7] fraction
   '(?:[ \\t]*(Z|([-+])([0-9][0-9]?)' + // [8] tz [9] tz_sign [10] tz_hour
-  '(?::([0-9][0-9]))?))?$');           // [11] tz_minute
+  '(?::?([0-9][0-9]))?))?$');           // [11] tz_minute
 
 function resolveYamlTimestamp(data) {
   if (data === null) return false;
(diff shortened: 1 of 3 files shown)


rlidwka:
> ISO-8601 dates can include a time offset from Zulu time, wither with a semi colon, such as "+05:00" or without a semi colon like "+0500."

That statement is as accurate as it is misleading.

ISO-8601 defines two different types of formats for each date representation:
 - extended format: `1985-04-12T10:15:30+04:00`
 - basic format: `19850412T101530+0400`

Punctuation is required in extended format and omitted in basic format.

We only support extended format because basic format isn't recommended for use in plain texts and can't be disambiguated from an integer in some cases.

> The issue isn't actually a Jekyll Issue, but is caused by js-yaml having slightly less support for ISO 8601 dates than other common YAML parsers.

Testing https://github.com/yaml/pyyaml:

```
$ python3
>>> import yaml
>>> yaml.safe_load('2001-12-14 21:59:43.10 -05:00')
datetime.datetime(2001, 12, 14, 21, 59, 43, 100000, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=68400)))
>>> yaml.safe_load('2001-12-14 21:59:43.10 -0500')
'2001-12-14 21:59:43.10 -0500'
```

One is recognized as a date, other one isn't.

PR #581, opened 2020-09-21
Set root export directory to lib/js-yaml/

xnuk:
This allows to import specific files only. Use case could be:

```js
const yaml = require('js-yaml/loader')
const FAILSAFE_SCHEMA = require('js-yaml/schema/failsafe')

yaml.safeLoad('hello-world: 1234.5678', { schema: FAILSAFE_SCHEMA })
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
   "bin": {
     "js-yaml": "bin/js-yaml.js"
   },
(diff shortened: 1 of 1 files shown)


puzrin:
No root hacks

PR #588, opened 2020-11-23
Fixed printable char detection for unicode > 0xFFFF

### LINKED ISSUE #587: dumper.js, why my emojis are escaped ?
Hi all,

The yaml specification says (https://yaml.org/spec/1.2/spec.html#id2770814):

> On output, a YAML processor must only produce acceptable characters. Any excluded characters must be presented using escape sequences. In addition, any allowed characters known to be non-printable should also be escaped. This isn’t mandatory since a full implementation would require extensive character property tables.

```
c-printable ::= #x9 | #xA | #xD | [#x20-#x7E]          /* 8 bit */
              | #x85 | [#xA0-#xD7FF] | [#xE000-#xFFFD] /* 16 bit */
              | [#x10000-#x10FFFF]                     /* 32 bit */
```

dumper.js follow the specification:
https://github.com/nodeca/js-yaml/blob/d8ba4022248136211675b3b29c522a53e6b5b1f7/lib/js-yaml/dumper.js#L186-L191

But the code that is calling isPrintable() use [charCodeAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt). This function will never return char > 0xFFFF so the supplementary planes range (0x010000 to 0x10FFFF) which include emojis will never be detected as printable.

https://github.com/nodeca/js-yaml/blob/d8ba4022248136211675b3b29c522a53e6b5b1f7/lib/js-yaml/dumper.js#L289-L293

I tested with [codePointAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt) and `Array.from(string)` to iterate the chars while managing surrogate pairs: emojis are not escaped as desired.

ysavary:
Fixes #587

### DIFF
--- a/test/issues/0369.js
+++ b/test/issues/0369.js
@@ -1,0 +1,10 @@
+'use strict';
+
+
+var assert = require('assert');
+var yaml = require('../../');
+
+
+test('should load astrals as codepoint', function () {
+  assert.deepEqual(yaml.safeLoad('"\\U0001F600"'), '😀');
+});
--- a/test/issues/0587.js
+++ b/test/issues/0587.js
@@ -17,8 +17,4 @@ suite('Should not escape emojis', function () {
   test('escape non-printables', function () {
     assert.strictEqual(yaml.safeDump('a😀\nb\u0001c'), '"a😀\\nb\\x01c"\n');
   });
-
-  test('load', function () {
-    assert.strictEqual(yaml.safeLoad('"\\U0001F600"'), '😀');
-  });
 });


puzrin:
Solved in more old-school way - with keeping older browsers support.

rlidwka:
Thanks for sending this PR.

It didn't quite work for us, since neither `Array.from()` nor `String.prototype.codePointAt()` are present in old browsers. So this issue was fixed slightly differently. So used a polyfill for `codePointAt`.

`Array.from()` isn't needed here at all, since it can be replaced with an iteration like this:

```js
/*
var chars = Array.from(string);
for (var i = 0; i < chars.length; i++) {
  var code = chars.codePointAt(0);
}
*/

for (var i = 0; i < string.length; i++) {
  var code = string.codePointAt(i);
  if (code >= 0x10000) i++;
}
```

... which does the same thing, but works faster.

ysavary:
Thank you @rlidwka for your fix !

PR #595, opened 2020-12-05
Convert the demo to use ESM

realityking:
Reduces the size of the demo bundle by ~15% (~90KB)

### DIFF
--- a/support/demo_template/.eslintrc.yml
+++ b/support/demo_template/.eslintrc.yml
@@ -1,0 +1,3 @@
+parserOptions:
+  ecmaVersion: 2015
+  sourceType: module
--- a/support/demo_template/base64.js
+++ b/support/demo_template/base64.js
@@ -28,8 +28,6 @@
 // [1]: http://lxr.mozilla.org/mozilla/source/extensions/xml-rpc/src/nsXmlRpcClient.js
 
 
-'use strict';
-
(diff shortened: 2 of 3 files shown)


puzrin:
Sorry, changes "just for fun" not accepted.

realityking:
@puzrin I'm not sure why you see this as "just for fun". It's a significant reduction of the bundle size of the demo, making it faster for anyone who uses it.

It's also a good testbed to verify the ESM module.

I don't feel strongly about it since it doesn't affect my downstream use of `js-yaml` but IMHO it's a nice win.

puzrin:
1.

> @puzrin I'm not sure why you see this as "just for fun". It's a significant reduction of the bundle size of the demo, making it faster for anyone who uses it.

IMO, since demo is not part of distributed code, any talks about demo size are pointless.

2.

> It's also a good testbed to verify the ESM module.

If you care about ESM verification - it's better to add explicit test.

3.

You suggested coding style change without asking maintainers if they wish/need that.

---

All that together looks in my eyes like change "just for fun". That doesn't mean you suggesed something bad. That means i don't have such fun with those changes an prefer spend efforts for for something else :)

realityking:
Fair enough if you don't care about the size of the demo. :) Thanks for the longer explanation.

PR #596, opened 2020-12-05
Clarify the impact of exports in the migration doc

realityking:
I'm not sure how common deep requires are Amon `js-yaml` users but I think it's good to call out that they're no longer possible. Especially since this only affects some Node.js versions.

### DIFF
--- a/migrate_v3_to_v4.md
+++ b/migrate_v3_to_v4.md
@@ -99,6 +99,9 @@ grep '\(^\|:\s\s*\)0[0-9][.0-9]*\s*$' *.yml
 
 If you have these in your yaml files anywhere, you should quote them (e.g. `0123456789` => `"0123456789"`).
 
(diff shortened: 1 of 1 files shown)


puzrin:
1. In upcoming 4.0 `Schema` is specially updated to reduce need of direct types use to zero.
2. Usually, such document are done as short as possible, to not demotivate anyone :).

If you have no samples of this problem directly for js-yaml migration, i'd suggest to postpone change until this happens at least once.

realityking:
Fair enough. I felt it was a good idea to include it due to the advice how to change deep requires right below this. For most Node users this would however not be possible at all.

Feel free to close it in case you don't want to merge it.

puzrin:
Let's close then. Will add as soon as anyone say been affected. Or may be option will be to add more dirs into exports.

PR #597, opened 2020-12-05
Fix broken build

realityking:
The branch doesn't build as the entry point file could not be found.

I'm not sure which naming/hierarchy you desire so I went with my usual pattern. Happy to adapt it if needed.


puzrin:
Thanks for PR, i have some claifications:

1. I do no accept `dist/` changes for security reasons. It worth remove that from PR.
2. It's better to make commit text more concrete, something like "update rollup paths after directory structure change" (https://github.com/nodeca/js-yaml/commit/cf8790b0a76d51f19665ba345bb5d9697020e5d0)
3. Some paths in `package.json` seems invalid - there are no `/lib/index.js`

If yo have no objection, i will fix that myself.

realityking:
1. Good to know. I'll omit these in the future
2. ✅ 
3. Note that the first commit in the PR moves `./index.js` to `./lib/index.js`

Feel free to fix this yourself, this way you can arrange things to your liking :)

puzrin:
I usually update `dist/` directly before npm publish. But if you need urgent update in `dev` branch - let me know.

PR #629, opened 2021-07-12
fix: remove support for underscore (`_`) in number parsing (fixes #627)

### LINKED ISSUE #627: Parsing numbers according to YAML 1.2
Based on the spec for YAML 1.2, underscores shouldn't be allowed in numbers:

https://yaml.org/spec/1.2/2009-07-21/spec.html#id2604185

However the latest `js-yaml` still allows them:

package.json
```
{
  "dependencies": {
    "js-yaml": "^4.1.0"
  }
}
```

test.js
```
const yaml = require('js-yaml');
const data = yaml.load(`
string: '1_2_3'
also_string: 1_2_3
`);
console.log(typeof data.string);
console.log(typeof data.also_string);
```

```
$ node test.js
string
number
```

Is this intended behaviour?  Thanks in advance.

david-mohr:
Parse numbers according to YAML 1.2 instead of YAML 1.1.

https://yaml.org/spec/1.2/2009-07-21/spec.html#id2604185


puzrin:
fixed by https://github.com/nodeca/js-yaml/commit/031ad079a65da5018cbe6280f3b9306e39699b5a

Issue #631, opened 2021-07-24
esm release tree shake is not pure


loynoir:
### Reproduce
```sh
echo 'export {} from "./dist/js-yaml.mjs"' | rollup
```

### Expected
output only LF

### Actual
output 1023 lines

puzrin:
I'm not familiar with details. Could you explain where to read and why functions should be marked manually (that looks strange)?

puzrin:
Google does not shows too much. This is interesting to read: https://github.com/d3/d3/issues/3076#issuecomment-644265124

Also:

- How standard is all that for libraries?
- Shouldn't we use [sideEffects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) option?
- All mentioned marks are `/*#__PURE__*/`, not `/*@__PURE__*/`. Where to read about difference?

loynoir:
(rollup pure)[https://rollupjs.org/guide/en/]
(esbuild pure)[https://esbuild.github.io/api/#pure]

both 2 seems equal.
I choose the one has more google result.

loynoir:
Set `sideEffects` should work too. Haven't tested.

# Scenario
`foo` depends on a bundlized library `bar`.
`bar` depends on `js-yaml`.

# Plan A
Pro: easist

Con: pure may lost, may pollute final output.

Set `sideEffects` within `js-yaml`.
`bar` got pure `js-yaml`.
But after bundlized, toplevel function call has no magic comments.
As `bar` itself might be un-pure, `foo` might un-pure library `bar`.


# Plan B
Pro: always pure, will never pollute final output.

Con: not easiest.

All `js-yaml` top level function call use magic comment.
`bar` got pure `js-yaml`.
Bundled `bar` still have magic comment, so it is pure too.
`foo` got pure `bar`.

puzrin:
@loynoir Sorry for delay with answer. I have no principal objections against marking required methods with comments, as you suggested.

Two more questions prior to move forward (sorry, if stupid, i'm not familiar with topic):

1. Why some functions of js-yaml stays unremoved? May be we could rewrite those (do some known recipes exist)?
2. How do you find, which functions  should be marked?

loynoir:
@puzrin 
I'm no expert. Someday I look some source code, found some magic comments, and did some search. (sorry, forget which project)

I have some thinks, but not sure if they are right.

1.
> Why some functions of js-yaml stays unremoved?
[...]

PR #632, opened 2021-07-24
Fix issue#631: esm release tree shake is not pure

loynoir:
make esm release tree shake pure

[issue631](https://github.com/nodeca/js-yaml/issues/631)


puzrin:
I'd suggest to remove `/dist/`  & `.gitignore` from PR, prior to continue. Current diff can not be verified.

PR #649, opened 2021-12-13
Add .parse alias and return undefined when loading nothing.

MistakingManx:
Added YAML.parse as an alias for YAML.load, and corrected a bug(?) where documents[0] returned "undefined" instead of undefined when calling YAML.load().

I did this before reading through the whole file, so I might make more changes later, but probably not.

Also pointing out that YAML.load("notafile") returns "notafile", which might be an issue too, but i dont want to look into it right now.


puzrin:
I see no reason to change api "just for fun". I'd suggest to create issues with details first to discuss.

PR #653, opened 2022-02-02
Add option to enforce flow syntax for arrays

td-sclemens:
This is adding an option to force the flow syntax for arrays when dumping to yaml string.


puzrin:
Solved in master (upcoming v5). Now AST can be transformed anyhow.

Issue #663, opened 2022-04-10
Module build failed: Module not found: "buffer"


shmolf:
### Full Output:
```
 WARN  Compiled with 1 warnings                                                                                                                                                                                                                                                   5:20:20 PM

Module build failed: Module not found:
"./node_modules/js-yaml/lib/js-yaml/type/binary.js" contains a reference to the file "buffer".
This file can not be found, please check it for typos or update it if the file got moved.

webpack compiled with 1 warning
Done in 114.26s.
```

### System:
```shell
yarn -v && npm -v && node -v
```
```
yarn: 1.22.15
npm: 8.1.2
node: v16.13.1
```

### Usage:
Using Webpack to bundle my assets for a website. This library is a dependency of a library used in the website.

shmolf:
As mentioned on the pull request (https://github.com/nodeca/js-yaml/pull/666), this is not to do with this library, but the environment to which it's used.

In my case, I recently upgraded from Webpack 4 to Webpack 5. In Webpack 5, polyfills for NodeJS APIs, are no longer supported (probably not wholly true but at least in part).

[Here's an article](https://viglucci.io/how-to-polyfill-buffer-with-webpack-5) describing how to polyfill `buffer` within a Browser-based application.

PR #666, opened 2022-05-01
fix(dependencies): add `buffer`

shmolf:
Related to issue I get when bundling my assets with WebPack.
Error occurs within Windows, Linux, and Docker containers.

```
Module build failed: Module not found:
"./node_modules/js-yaml/lib/js-yaml/type/binary.js" contains a reference to the file "buffer".
This file can not be found, please check it for typos or update it if the file got moved.
```

Addresses issue https://github.com/nodeca/js-yaml/issues/663

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7068 @@
+{
+  "name": "js-yaml",
+  "version": "4.1.0",
+  "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


puzrin:
Please investigate bundler docs how to auto-include node's built-ins, or use files from `/dist`.

PR #677, opened 2022-07-06
Fix package.json exports

victorenator:
https://nodejs.org/dist/latest-v18.x/docs/api/packages.html#subpath-exports

Nested levels are not supported.


puzrin:
Not actual

PR #683, opened 2022-09-14
Allow blockquoted tabs

### LINKED ISSUE #682: Strings with tabs and newlines can't be serialized as block quotes
The dumper code (in `chooseScalarStyle`) checks in a couple places to see if any characters in the string to be written are non-printable, and if so only allows double-quoted output. This is mostly fine, except tabs are explicitly excluded from the "printable" set (there's a comment implying it's intentional). That means that any block quote with tabs is unconditionally rendered as a single-line double-quoted string.

For example, loading+dumping this:

```yaml
value: |-
  a
  	b
```

results in this:
```yaml
value: "a\n\tb"
```

obviously this example is trivial and no big deal, but with a very large string that has embedded tabs (for example, an example program embedded in a value of a yaml block), you can imagine it gets pretty ugly.

For my needs it'd be enough to change the `isPrintable` checks in `chooseScalarStyle` to also accept tabs (0x09)~~, but I think there's a pretty reasonable case to be made for carriage returns and non-breaking spaces too.~~ It turns out I also need carriage returns and newlines, so I've patched that as well in my PR.

beckjake:
Allow tabs, newlines, and carriage returns in blockquoted strings + add a test. Fixes #682


puzrin:
https://github.com/nodeca/js-yaml/blob/master/docs/scalar_styling.md

Fixed in new styling API

PR #687, opened 2022-10-10
fix: resolving anchors across documents

katzer:
```yml
--- &default
foo: bar
---
<< : *default
bar: baz
```


puzrin:
Closing as invalid - anchors are local for the document, by spec.

PR #699, opened 2023-04-02
ci: Update to currently supported Node.js versions

nschonni:
Dropped the EOL Node 12 from the testing, and added the newer versions to the matrix.
Bumped the GitHub actions to the latest versions as well


puzrin:
Solved in another commit

PR #714, opened 2023-11-30
Adding a Native Multi Yaml Parser

AdrianMartinezCodes:
Having multiple yamls defined in one file has become common practice in recent years. It would be useful if this was a native api the library provided for users rather than defining their own iterable. This would also return them in one array rather than separate returns.

### DIFF
--- a/test/samples-common/construct-multi-yaml.yml
+++ b/test/samples-common/construct-multi-yaml.yml
@@ -1,0 +1,23 @@
+---
+name: Example Document 1
+items:
+  - part: A
+    quantity: 10
+  - part: B
+    quantity: 5
+
+---
+name: Example Document 2
+info:
+  description: This is a test document
+  version: 2
+values:
+  - type: alpha
+  - type: beta
+  - type: gamma
+
+---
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -21,6 +21,7 @@ module.exports.CORE_SCHEMA         = require('./lib/schema/core');
 module.exports.DEFAULT_SCHEMA      = require('./lib/schema/default');
 module.exports.load                = loader.load;
 module.exports.loadAll             = loader.loadAll;
+module.exports.loadMultiYaml       = loader.loadMultiYaml;
```
rlidwka, line 24: Thank you for the pull request. However the new `loadMultiYaml` seem to do exactly what `loadAll` function above does, and it's not immediately clear why we need both.

AdrianMartinezCodes: You're correct, I misread the documentation. I think a better idea would be to showcase the example of outputting the multi-document sources. E.g. the case where we don't apply the iterator.

I was under the belief the loadAll() function was only loading them in one at time(not storing to an array directly).

PR #716, opened 2024-01-22
fix: tag-handle may include digits in directive

mighty1231:
### What does it fix?

normal tag directive - `%TAG !taghandle! tag:yaml.org,2002:`

test case `%TAG !12345! tag:yaml.org,2002:`

- **currently** `js-yaml` throw error during parsing tag directive
- **expected** it should not throw error

reference: [c-tag-handle](https://yaml.org/spec/1.1/#c-tag-handle)

- "c-tag-handle" may be "c-named-tag-handle"
- c-named-tag-handle ::= "!" ns-word-char + "!"
- [ns-word-char](https://yaml.org/spec/1.1/#ns-word-char) defined as one of ns-dec-digit, ns-ascii-letter or "-"
- and finally `ns-dec-digit` is `[0-9]`

My references is from yaml 1.1, but these things have not been changed in yaml 1.2 spec


puzrin:
Solved in master (upcoming v5), as part of global sync with spec syntax

PR #741, opened 2026-03-12
Document `listener` option

joshkel:
Add description for 'listener' option in README. This was implemented in #248 but never documented.


scutuatua-crypto (approved):

puzrin:
Listener support in current state is "unofficial". It was added as quick-kludge, without any serious architecture in mind. I don't like to take responsibility about this solution and it's maintenance. You can use it if you wish, but on your own risk. It's intended to keep it out of main docs.

You can dig previous talks - this was declared explicit prior to accept PR with implementation.

PR #747, opened 2026-05-09
fix(float): reject out-of-range values that would parse to Infinity

### LINKED ISSUE #737: Can't round-trip out-of-range floating point
**Describe the bug**
When parsing a document with an unquoted string that "looks like" a number, the parser will use the float tag, even though the value cannot be represented by the JavaScript `number` and ends up getting stored a +Inf or -Inf.

**To Reproduce**
Steps to reproduce the behaviour:
```js
> require("js-yaml").load("gitsha: 61e9540")
{ gitsha: Infinity }
```

**Expected behaviour**
Like some other YAML loaders (go-yaml comes to mind), an untagged, unquoted value should only be interpreted as a float if it can be safely stored as such in the native type. 

**Versions (please complete the following information):**
- NodeJS 24.12.0
- `js-yaml`: 4.1.1

**Additional context**
[The YAML 1.2 spec is pretty clear:](https://github.com/yaml/yaml-spec/blob/main/spec/1.2/markdown/spec.md#definition-6)
> Some languages (such as Perl) provide only a "number" type that allows for both integer and floating-point values. A YAML [processor] may use such a type for floating-point numbers, **as long as they round-trip properly.**

Jah-yee:
## Summary

Values like `61e9540` (e.g., a git SHA) are parsed as floats and become `Infinity`, breaking round-trip as required by YAML 1.2 spec.

**Before:**
```js
yaml.load('gitsha: 61e9540') // { gitsha: Infinity }
```

**After:**
```js
yaml.load('gitsha: 61e9540') // { gitsha: '61e9540' }
```

## Fix

In `resolveYamlFloat()`, pre-check if `parseFloat(data)` would produce a finite number. Explicit `.inf`/`.nan` are still allowed. All other behavior is unchanged.

## Changes

- **lib/type/float.js**: Added finite-number check in `resolveYamlFloat`

## Testing

```
61e9540 -> '61e9540' (string, not Infinity)
1e308 -> 1e308 (still finite float)
1e309 -> '1e309' (string, would otherwise be Infinity)
.inf -> Infinity (still allowed)
.nan -> NaN (still allowed)
3.14 -> 3.14 (normal floats unchanged)
```

All 64 existing unit tests pass.

## Fixes

Fixes #737

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -902,6 +902,13 @@ function readBlockScalar(state, nodeIndent) {
       continue;
     }
 
+    // Zero-indentation block scalar is not allowed.
+    // If textIndent is still 0 at this point, it means no explicit indentation
+    // indicator was given and no indentation was detected in content lines.
+    if (!detectedIndent && textIndent === 0) {
+      throwError(state, 'missing indentation for block scalar');
+    }
+
     // End of the scalar.
     if (state.lineIndent < textIndent) {
 
--- a/lib/type/float.js
+++ b/lib/type/float.js
@@ -24,6 +24,15 @@ function resolveYamlFloat(data) {
     return false;
   }
 
+  // Reject values that would parse to Infinity/NaN (except explicit .inf/.nan)
+  // This ensures round-trip: out-of-range ints like "61e9540" become strings
+  var value  = data.replace(/_/g, '').toLowerCase();
+  var sign   = value[0] === '-' ? -1 : 1;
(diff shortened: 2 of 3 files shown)


Jah-yee:
Update: squashed commits into a single commit. 

Changes consolidated:
- lib/type/float.js: reject out-of-range float values (fixes #737)
- lib/loader.js: throw error on zero-indentation block scalar

Ready for review! 🙏

PR #755, opened 2026-05-27
chore: update package.json structure

nschonni:
Ran `npm init -y` to get some of the old field structures into the newer format. Dropped a few of the new fields.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,12 +10,31 @@
   ],
   "author": "Vladimir Zapparov <dervus.grim@gmail.com>",
   "contributors": [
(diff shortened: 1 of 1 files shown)


puzrin:
Looks like a change just for fun, without actual need. Please, don't touch project configs. Create an issue with explanations if something really needs care.

nschonni:
Fair enough. If you look at https://docs.npmjs.com/cli/v11/configuring-npm/package-json the fields I touched changed the format (I think in NPM 6 or before)

puzrin:
I think in terms of "added value". IMHO, the file just gets bigger for no useful reason.

PR #757, opened 2026-06-01
Preserve object keys in !!pairs instead of stringifying to '[object Object]'

### LINKED ISSUE #481: First item in a !!pair is unnecessarily converted into string
```yaml
pairs:
  Block tasks: !!pairs
    - { test: 123 }: { test: 123 }
    - { test: 456 }: { test: 456 }
```

```json
{ pairs: 
   { 'Block tasks': 
      [ [ '[object Object]', { test: 123 } ],
        [ '[object Object]', { test: 456 } ] ] } }
```

In normal maps `{ test: 123 }` is turned into string because javascript doesn't support objects as keys (Maps aside, as we don't support those).

But pairs are turned into arrays, which *can* have objects inside. So stringification is unnecessary there.

*somewhat related to #169, but this is an issue that can be fixed without rewriting everything*

viraatdas:
Fixes #481.

A `!!pairs` item whose key is a flow mapping (an object) was stringified to `'[object Object]'`, because by the time the `!!pairs` constructor runs, the loader has already converted the key to a plain-object string key.

This captures the original key node in the loader for object/sequence keys (kept in a `WeakMap` side-table so it is garbage-collected with the mapping and does not change behavior for normal mappings), and the `!!pairs` constructor recovers it. Scalar, numeric, and timestamp keys keep their existing string behavior.

Added `test/core/issues/0481.js` covering object keys, scalar keys, numeric keys, flow style, and timestamp keys. Full suite passes.

This is a non-trivial change to the loader; happy to gate the metadata tracking more narrowly (e.g. only when a `!!pairs` tag is involved) if you prefer a lighter touch.

Prepared with AI assistance.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -4,6 +4,7 @@ const common = require('./common')
 const YAMLException = require('./exception')
 const makeSnippet = require('./snippet')
 const DEFAULT_SCHEMA = require('./schema/default')
+const pairMetadata = require('./pair-metadata')
 
 const _hasOwnProperty = Object.prototype.hasOwnProperty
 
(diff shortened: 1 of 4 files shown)


puzrin:
That's too hacky.

https://github.com/nodeca/js-yaml/commit/42dea284430b1077f95a326f14db422b19bedecc

In master (upcoming v5), everything should work if you use the `realMap` tag (Map-based) instead of the default one.

PR #758, opened 2026-06-03
Quote scalars that look like a document-separator line

sarathfrancis90:
### Problem

`yaml.dump()` emits a plain scalar for strings that the parser then reads as a document separator, breaking the round trip:

```js
yaml.load(yaml.dump('...'))    // => null   (expected '...')
yaml.load(yaml.dump('... x'))  // => throws "expected a single document in the stream, but found more"
```

The dumper guarded only the exact strings `'---'` / `'...'`, but the loader's `testDocumentSeparator()` treats `---`/`...` at the start of a line **followed by whitespace or end-of-line** as a separator. So `... x`, `--- x`, etc. slipped through as plain scalars.

### Fix

Mirror the loader's rule in the dumper (`/^(---|\.\.\.)(\s|$)/`) so any scalar whose first line is a document-separator line is quoted. Strings that merely resemble markers (`...x`, `....`, `x...`) stay plain — no over-quoting.

Added round-trip assertions for `...`, `---`, `... x`, `--- x` and negative cases. `npm test` (lint + core + build) passes.

### DIFF
--- a/lib/dumper.js
+++ b/lib/dumper.js
@@ -176,6 +176,14 @@ function testImplicitResolving (state, str) {
   return false
 }
 
+// A plain scalar equal to "---" or "..." at the start of a line is parsed as a
+// directives-end / document-end marker (see [206] c-forbidden in the spec, and
+// testDocumentSeparator() in the loader), so such a string must be quoted to
+// survive a round trip.
+function isDocumentSeparator (string) {
+  return string === '---' || string === '...'
+}
+
 // [33] s-white ::= s-space | s-tab
 function isWhitespace (c) {
   return c === CHAR_SPACE || c === CHAR_TAB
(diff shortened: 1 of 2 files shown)


puzrin:
I did a big refactoring for v5, and this PR is out of sync after the branch merge.

1. https://github.com/nodeca/js-yaml/commit/6cf374ee84f51d78a8286bbcbbd3ac2628153927 this quotes `---`, `...`, `- value` and `? value` cases.
2. [Quote document markers followed by whitespace](https://github.com/nodeca/js-yaml/commit/704b25d29229c52e66b824bfc1908d8f3ba3a45e) this quotes `--- foo` / `... foo`

Sorry for not waiting for your updates, but big thanks for PR, as it encouraged me to find more bugs and make fixes.

sarathfrancis90:
No worries at all — glad the PR nudged those v5 fixes out. Your handling covers more ground than my single document-separator patch did (the `- `/`? ` and `--- foo` whitespace cases especially). Thanks for the pointers to the commits.

PR #763, opened 2026-06-19
fix: quote dumped strings that YAML 1.1 reads as underscore numbers

### LINKED ISSUE #760: Since version 4.2.0 string is dumped as YAML number
Since version 4.2.0 strings like `"0_30"` or `"30_45"` are dumped as number.

```
const ymldoc = yaml.dump({test: '0_30'});
console.log(ymldoc);
```

generates following yaml `test: 0_30`.

Up to version 4.1.1 the YAML output was `test: '0_30'`, as expected.

Additional Note:

If  the `{forceQuotes: true}` option is used, the rendered yaml is `test: '0_30'`.

vjsai:
Closes #760

## Problem

`yaml.dump({ test: '0_30' })` emits `test: 0_30` **unquoted**:

```js
yaml.dump({ test: '0_30' })
// before: "test: 0_30\n"
```

YAML 1.1 parsers (PyYAML, libyaml, Psych) and js-yaml ≤ 4.1.1 resolve `0_30` as the integer `30`, so the unquoted output is no longer round-trip safe across parsers — the string silently becomes a number when read elsewhere.

This regressed as a **side effect** of #627. That commit correctly made the loader YAML-1.2-compliant: `lib/type/int.js` / `float.js` no longer resolve numbers with underscores. As a consequence, the dumper's `testImplicitResolving` no longer flags these strings, so `chooseScalarStyle` emits them plain.

**This PR does not revert #627.** The loader change is intentional and correct, and is left untouched. The fix is entirely in the dumper: it restores cross-parser round-trip safety by *quoting* a plain string that a YAML 1.1 parser would read as a number.

## Fix

Add a dumper guard mirroring the existing `DEPRECATED_BASE60_SYNTAX` guard in `writeScalar` (it sits in the same `!noCompatMode` block, alongside the deprecated-boolean and base-60 guards). The new `DEPRECATED_UNDERSCORE_NUMBER_SYNTAX` regex + `isDeprecatedUnderscoreNumber` predicate reproduce the **pre-#627 loader semantics** exactly: decimal int/float, hex/octal/binary, with no trailing underscore, reducing to a finite number. So only the forms a YAML 1.1 parser actually resolved as numbers get quoted.

```js
yaml.dump({ test: '0_30' })
// after: "test: '0_30'\n"
```

Forms YAML 1.1 never resolved stay plain (no over-quoting):

| input | before | after |
|---|---|---|
| `'0_30'` | `0_30` | `'0_30'` |
| `'1_000'` | `1_000` | `'1_000'` |
| `'0x1_F'` | `0x1_F` | `'0x1_F'` |
| `'_30'` | `_30` | `_30` (unchanged) |
| `'30_'` | `30_` | `30_` (unchanged) |
| `'hello_world'` | `hello_world` | `hello_world` (unchanged) |

`{ forceQuotes: true }` already quoted these; this addresses the *implicit* quoting decision. `{ noCompatMode: true }` still opts out, consistent with the existing base-60 / boolean compat behavior.

## Test

Added `test/core/issues/0760.js` (auto-required by the issues runner). Verified fails-before / passes-after: reverting just the dumper change makes it fail with `actual: 'test: 0_30\n'` vs `expected: "test: '0_30'\n"`; re-applying makes it pass.

## Full test suite (`npm test`)

```
> npm run lint  → 0 errors (1 pre-existing warning in loader.js, unrelated)

> test:core
ℹ tests 304
ℹ suites 26
ℹ pass 303
ℹ fail 0
ℹ skipped 1

> test:build
ℹ tests 6
ℹ pass 6
ℹ fail 0
```

---

Prepared with AI assistance (Claude Code), reviewed by the author.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+### Fixed
+- Quote dumped strings that YAML 1.1 parsers would resolve as numbers with
+  underscores, keeping output round-trip safe across parsers, #760.
+
+
 ## [4.2.0] - 2026-06-01
 ### Added
 - Added `docs/safety.md` with notes about processing untrusted YAML.
--- a/lib/dumper.js
+++ b/lib/dumper.js
@@ -58,6 +58,15 @@ const DEPRECATED_BOOLEANS_SYNTAX = [
 
 const DEPRECATED_BASE60_SYNTAX = /^[-+]?[0-9_]+(?::[0-9_]+)+(?:\.[0-9_]*)?$/
 
+// Underscores in numbers used to be resolved as numeric scalars (YAML 1.1), but
+// no longer are (see #627). These forms still resolve as numbers in YAML 1.1
+// parsers (PyYAML, libyaml, Psych) and in js-yaml <= 4.1.1, so a plain string
+// matching one of them would not round-trip cross-parser and must be quoted.
+// Mirrors the pre-#627 loader: decimal int/float, hex/octal/binary, no trailing
+// underscore. The parseFloat check rejects degenerate forms (e.g. `._e0`).
+const DEPRECATED_UNDERSCORE_NUMBER_SYNTAX =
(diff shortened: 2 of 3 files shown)


puzrin:
I think such things can be endless. Take a look at the `dev` branch with upcoming v5. It does quoting in a much better and more predictable way. It's dumper uses yaml 1.1 + some missed things like `0oXXX` from 1.2 core schema, by default. Without hardcoding. And anyone can select another (or their own) schema if an old mess is not needed. So the question of compatibility is solved once and for all.

I hope to release it in a week. IMHO, it's better to keep v4 stale.

Issue #280, opened 2016-04-15
Loader must throw an error on zero-indentation block scalar


dervus:
Now it doesn't and consumes rest of the input as valid part of block scalar.

Jah-yee:
## PR Available ✨

PR #744 addresses this exact issue. Here's a summary of the fix:

**Problem:** The loader silently accepts zero-indentation block scalars (which are ambiguous and invalid per YAML spec), leading to unexpected behavior.

**Solution:** Added a validation check that throws a descriptive error when a block scalar marker (| or >) is followed by a line with no indentation.

**Changes:**
- Added a  check for indentation before processing block scalars
- Error message clearly explains the problem
- Only 1 commit, minimal diff

The PR is currently open at: https://github.com/nodeca/js-yaml/pull/744

Would appreciate a review! 🙏

Jah-yee:
## PR Available

PR #744 addresses this issue about zero-indentation block scalars.

**Problem:** Zero-indentation block scalars (| or > followed by unindented content) are ambiguous/invalid per YAML spec.

**Solution:** Added validation to throw a descriptive error in this case.

**PR:** https://github.com/nodeca/js-yaml/pull/744

Would appreciate a review! 🙏

puzrin:
Reopen since PR reverted

puzrin:
It would be clearer to check for finiteness first and apply a part of the validation regexp if not.

I'm sorry, but too many microchanges require more attention than I can give now. Let me do a normal planning and a bulk update as intended. If this pr will stay actual after - I will return to it.

puzrin:
W4TN, M7A3, DWX9 of yaml test suite allow this behaviour. Probably, need roll back it for v4.

In v5 works as spec tests say.

puzrin:
Opening as v4 memo, to roll back.

PR #765, opened 2026-06-19
Fix dump of whole-document scalar made only of line breaks

chatman-media:
When the input is a string consisting **only of line breaks**, `dump()` produced output that the loader cannot parse, so the round trip throws:

```js
const yaml = require("js-yaml");

yaml.load(yaml.dump("\n"));     // throws: missing indentation for block scalar (3:1)
yaml.load(yaml.dump("\n\n"));   // throws
```

`dump("\n")` emits:

```
|+

```

i.e. a literal block scalar with the `+` (keep) chomping indicator but **no indentation indicator**, followed by blank lines. A block scalar made up entirely of line breaks has no non-empty line, so the parser cannot auto-detect the content indentation. When the scalar is nested under a mapping/sequence key the parent node pins the indentation and it parses fine, but as a whole document there is no parent, so the loader rejects it via the same guard added in `4ed3479` ("Reject top-level block scalars without content indentation", #280).

### Fix

`blockHeader()` now emits an explicit indentation indicator for top-level scalars that consist only of line breaks. The indicator width is irrelevant in this case (there is no content to indent), so it is clamped to a single digit to stay within the spec's allowed range. Output for nested scalars is unchanged.

```js
yaml.dump("\n");    // "|2+\n\n"   ->  loads back as "\n"
yaml.dump("\n\n");  // "|2+\n\n\n" ->  loads back as "\n\n"
```

### Tests

`test/core/units/dump-scalar-styles.js` previously asserted the broken `|+` output (it only checked the dump string, never the round trip). Updated to the corrected output and added round-trip assertions (`yaml.load(yaml.dump("\n")) === "\n"`). The test fails on `master` without the dumper change and passes with it.

Verified with a 400k-case dump→load fuzz over short strings of YAML-significant characters: parse-throw failures drop from 1685 to **0**, with no change to any other round-trip result. Full suite (`npm test`: lint + core + build) is green.

### DIFF
--- a/lib/dumper.js
+++ b/lib/dumper.js
@@ -427,10 +427,10 @@ function writeScalar (state, string, level, iskey, inblock) {
       case STYLE_SINGLE:
         return "'" + string.replace(/'/g, "''") + "'"
       case STYLE_LITERAL:
-        return '|' + blockHeader(string, state.indent) +
+        return '|' + blockHeader(string, state.indent, level === 0) +
           dropEndingNewline(indentString(string, indent))
       case STYLE_FOLDED:
-        return '>' + blockHeader(string, state.indent) +
+        return '>' + blockHeader(string, state.indent, level === 0) +
           dropEndingNewline(indentString(foldString(string, lineWidth), indent))
       case STYLE_DOUBLE:
         return '"' + escapeString(string, lineWidth) + '"'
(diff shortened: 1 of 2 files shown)


puzrin:
Should be solved in master (upcoming v5)

PR #775, opened 2026-07-22
Fix !!timestamp resolution for years 0000-0099

sarathfrancis90:
While round-tripping `Date` values through `YAML11_SCHEMA` I hit a case where a date in years 1–99 throws on reload:

```js
const d = new Date(0); d.setUTCFullYear(50, 0, 1)
load(dump(d, { schema: YAML11_SCHEMA }), { schema: YAML11_SCHEMA })
// -> throws: cannot resolve a node with !<tag:yaml.org,2002:timestamp> explicit tag
```

And loading a valid low-year timestamp directly silently yields a string instead of a `Date`:

```js
load('0001-01-01', { schema: YAML11_SCHEMA }) // -> '0001-01-01' (string), expected a Date
```

Root cause: `resolveYamlTimestamp` builds the date with `Date.UTC(year, …)`, which applies JavaScript's legacy two-digit-year remap (0–99 → 1900–1999). The Date is then off by ~1900 years, so the calendar-normalization guard (`getUTCFullYear() !== year`) rejects the value and it falls back to a plain string. During `dump` that same rejection forces an explicit `!!timestamp` tag onto the scalar, which then can't resolve on load — hence the throw.

Fix: restore the intended year with `setUTCFullYear(year, month, day)` before the validity check. Passing month/day re-applies the day in the context of the real year, so invalid calendar dates like `0001-02-29` are still rejected.

Tested with `npm test` (all green). Extended `test/core/tags/timestamp.test.mjs` with low-year load cases, a leap day (`0004-02-29`), rejection of an invalid low-year date, and a dump → load round-trip.

### DIFF
--- a/src/tag/scalar/timestamp.ts
+++ b/src/tag/scalar/timestamp.ts
@@ -27,6 +27,10 @@ function resolveYamlTimestamp (source: string) {
   // Date-only form (`YYYY-MM-DD`) has no time captures.
   if (!match[4]) {
     const date = new Date(Date.UTC(year, month, day))
+    // Date.UTC() maps years 0-99 into 1900-1999; restore the intended year so
+    // low four-digit years (e.g. 0001) resolve correctly. Passing month/day
+    // re-checks the leap day against the real year.
+    if (year >= 0 && year <= 99) date.setUTCFullYear(year, month, day)
     // Reject dates that JS would normalize, e.g. 2023-02-29 -> 2023-03-01.
     if (date.getUTCFullYear() !== year || date.getUTCMonth() !== month || date.getUTCDate() !== day) {
       return NOT_RESOLVED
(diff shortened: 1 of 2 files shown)


puzrin:
I understand the problem, but I don't like the implementation with a duplicated check for the year subrange. IMHO, it's worth creating a helper

```js
function makeUtcDate (year, month, day, hour = 0, minute = 0, second = 0, fraction = 0) {
  const date = new Date(Date.UTC(year, month, day, hour, minute, second, fraction))

  // Date.UTC() treats years 0..99 as 1900..1999. Restore the parsed YAML year
  // before validating calendar normalization, e.g. reject 0001-02-29.
  date.setUTCFullYear(year, month, day)

  return date
}
```

That's more readable and "straight".

---

Tests are bloated. It's worth just adding a couple of good samples to the existing one (`0001-01-01`, `0050-06-15T12:30:00Z`) and `0001-02-29` to the existing test of rejected examples.

Issue #538, opened 2020-03-07
loading non-string keys with JSON_SCHEMA should error?


notEthan:
hi nodeca friends. I am trying to restrict the documents a tool can parse to those compatible with JSON. for values not supported by JSON, js-yaml will raise `YAMLException`, which is what I would expect. but for keys, it silently stringifies them.

```javascript
yaml.safeLoad("1: 'one'", {schema: yaml.JSON_SCHEMA})
// returns { '1': 'one' }
```

is it possible to detect and reject documents which do not have JSON compatible keys using js-yaml?

notEthan:
I'm realizing, actually, looking through YAML's [JSON schema](https://yaml.org/spec/1.2/spec.html#id2803231), it doesn't actually say anything about mapping keys, meaning the failsafe definition would apply ("YAML places no restrictions on the type of keys; in particular, they are not restricted to being scalars"). despite that not being compatible with JSON.

so this issue isn't really about JSON_SCHEMA, I guess. maybe a more generic option to reject key types which cannot be represented in JS, rather than stringifying, is appropriate?

dota17:
> hi nodeca friends. I am trying to restrict the documents a tool can parse to those compatible with JSON. for values not supported by JSON, js-yaml will raise `YAMLException`, which is what I would expect. but for keys, it silently stringifies them.
> 
> ```js
> yaml.safeLoad("1: 'one'", {schema: yaml.JSON_SCHEMA})
> // returns { '1': 'one' }
> ```

as i know, even if for values not supported by JSON, js-yaml will not raise `YAMLException`,either. Can you show your error value case ?

notEthan:
the readme notes in section [safeLoad](https://github.com/nodeca/js-yaml/#safeload-string---options-) some cases where `YAMLException` is raised. I'm trying to detect and reject documents which do not have JSON compatible keys - `YAMLException` might be the correct vehicle to do that, or maybe a different control flow would be more appropriate.

puzrin:
js-yaml does NOT throw exception for valid yaml. Because valid yaml must be loaded. But some types are loaded different https://github.com/nodeca/js-yaml#supported-yaml-types. Keys are stringified somehow.

Restriced schema will cause just more plain scalars. IMO your question is not specific to this package

dota17:
use [custom type](https://github.com/nodeca/js-yaml/wiki/Custom-types) to detect unvalid keys and throw Error.
```js
[...]

Issue #329, opened 2017-02-17
Support YAML maps and sets to ES6/ES2015 Maps and Sets


js-choi:
js-yaml current converts YAML `!!map` and `!!set` data structures to JavaScript objects with enumerated keys. It would make sense and be useful to optionally convert YAML `!!map`s and `!!set`s to ES6 / ECMAScript 2015 `Map`s and `Set`s instead; they would also support non-string keys and be convertible back to YAML `!!map`s and `!!set`s.

puzrin:
Seems dupe of #169.

I this request "just for fun" of you have real use-cases when maps/sets required?

js-choi:
It's not a very complicated use case. My codebases (many of which involve wide-breadth tree manipulation) use Maps and Sets already, and I would like to be able to serilaize and deserialize those data into YAML without allocating unnecessary objects returned by js-yaml. I would also use the YAML to interchange the data with old non-JS programs, which sometimes do use non-string keys.

puzrin:
It doesn't matter if use case is complicated or not. It make sense if use case is valuable for others and follows yaml spirit. I understand that it's technically possible to use maps & sets in DB. But in general, it's not a great idea to use yaml for DB serialization. Yaml is intended first to be human readable.

Also, as you can see from 169, there are bad side effects with maps.

In short - situation with maps/sets is known, but there are no plans to modify current behaviour in nearest future.

js-choi:
My use case does not involve YAML for entire databases, but rather interchanging JavaScript objects, many of which are Maps and Sets, between programs and serializing them into a format that can be read and edited by human users. These are primarily for configuration files with complex data, such as specifying chains of tasks in syntactic data structures that form a DSL. These configuration files need to be able to distinguish sequences, sets, and maps. There are similar use cases for general graph serialization. In other programming languages such as Python and Clojure, I've used YAML as well as [EDN](edn-format.org) and [Transit](https://github.com/cognitect/transit-format) for similar uses.

Having said all that, if adding support for Maps and Sets would induce side effects that affect existing users of this library, then I understand being adverse to doing so. I had assumed that it would be a matter of adding an opt-in mode while keeping previous behavior the same. Thank you for considering and responding to this issue; I will find another solution.

puzrin:
Problem is that yaml makes no difference between hashes (JS objects with simple keys) and maps. Option can only force to use maps everywhere, not in specific cases. But that makes result ugly for JS, as discussed in #169.

[...]

PR #776, opened 2026-07-23
Allow a local key to override a merged non-string key

spokodev:
### Problem

A merge key (`<<`) folds another mapping's keys into the target and records them
as *overridable*, so a later explicit pair may override a merged key without a
`duplicated mapping key` error (the canonical YAML merge "Override" behavior).

For the default object-based map, that tracking breaks on **non-string** keys:

```js
const { load, CORE_SCHEMA, mergeTag } = require('js-yaml');
const schema = CORE_SCHEMA.withTags(mergeTag);

load('base: &a {10: x}\nm:\n  <<: *a\n  10: y', { schema });
// throws YAMLException: duplicated mapping key
```

A string key in the same position works fine (`ten:` → `{ten: 'y'}`). Non-string
map keys are ordinary in YAML (`on/off/yes/no`, port numbers, numeric enum keys).

### Root cause

The overridable set is populated from `keys()`, which for the object map is
`Object.keys()` — always **strings** (`"10"`). The duplicate check then queries
that set with the **resolved** key (the number `10`):

```ts
if (!state.json && frame.tag.has(frame.value, key) && !frame.overridable?.has(key)) {
  throwError(state, 'duplicated mapping key')
}
```

`frame.tag.has` coerces (`String(key)`) so it correctly sees the duplicate, but
`Set.has` uses SameValueZero, so `"10" !== 10` — the override exemption misses
and the pair is rejected. `realMapTag`/`setTag` store keys losslessly, so they
were never affected, which pinpoints the object-map string coercion as the cause.

### Fix

Object-based maps already normalize keys to their string form in
`addPair`/`has`/`get`. Give a mapping tag a `normalizeKey` (defaulting to
identity, so `Map`/`Set` tags are unchanged) and key the overridable set through
it, so both sides use the same key identity.

- Genuine duplicate keys without a merge still raise (`{10: x, 10: y}` throws).
- String-key overrides are unchanged.

### Test

Added a `tags/merge` case where a merged `10:`/`true:` is overridden by a local
pair. Fails on `main` (`duplicated mapping key`), passes with the fix. Full
`test/core` suite green (284 pass / 0 fail), `type-check` and `lint` clean.

### DIFF
--- a/src/parser/constructor.ts
+++ b/src/parser/constructor.ts
@@ -264,7 +264,7 @@ function mergeKeys (state: ConstructorState, frame: MappingFrame, source: unknow
 
     const err = frame.tag.addPair(frame.value, sourceKey, sourceTag.get(source, sourceKey))
     if (err) throwError(state, err)
-    ;(frame.overridable ??= new Set()).add(sourceKey)
+    ;(frame.overridable ??= new Set()).add(frame.tag.normalizeKey(sourceKey))
   }
 }
 
@@ -295,13 +295,13 @@ function addMappingValue (state: ConstructorState, frame: MappingFrame, key: unk
     return
   }
 
-  if (!state.json && frame.tag.has(frame.value, key) && !frame.overridable?.has(key)) {
+  if (!state.json && frame.tag.has(frame.value, key) && !frame.overridable?.has(frame.tag.normalizeKey(key))) {
     throwError(state, 'duplicated mapping key')
   }
 
   const err = frame.tag.addPair(frame.value, key, value)
   if (err) throwError(state, err)
-  frame.overridable?.delete(key)
+  frame.overridable?.delete(frame.tag.normalizeKey(key))
 }
 
 function addValue (state: ConstructorState, value: unknown, tag: AnyTag) {
--- a/src/tag.ts
+++ b/src/tag.ts
@@ -58,6 +58,11 @@ interface MappingTagDefinition<Carrier = unknown, Result = Carrier> {
   has: (carrier: Carrier, key: unknown) => boolean
   keys: (result: Result) => Iterable<unknown>
   get: (result: Result, key: unknown) => unknown
+  // Canonical identity of a key, used when tracking merge-overridable keys.
+  // For object-based maps a resolved key (e.g. the number 10) and the string
+  // form it is stored under ("10") are the same key, so they must normalize
+  // alike; real Map/Set keys are already identity.
+  normalizeKey: (key: unknown) => unknown
   finalize: (carrier: Carrier) => Result
   carrierIsResult: boolean
   identify: IdentifyFn | null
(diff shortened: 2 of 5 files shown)


puzrin:
I understand the problem, but not sure about implementation.

The enforced key stringification for JS objects is language-specific behavior, not a "normal" one. So:

1. Introducing a "normalize" method in the collection tag is very confusing.
2. This is a specific tag-level defect. It would be nice to avoid spreading it into the upper level, if possible.

 IMHO, worth discussing alternatives in the issue first.

Also, note about the topic name - it's about scalar keys only. Non-scalar keys (complex objects) are usually not supported at all (PyYAML, Go) or are potentially vulnerable to DoS attacks.

spokodev:
Yeah, that's fair. And you're right it's only scalar keys — the map's `addPair` already rejects complex ones.

The reason it ends up in the constructor is that the `overridable` set lives there, not in the tag. For the plain-object map a merged key comes back from `keys()` as `"10"`, but the explicit override key is the resolved number `10`, and they only line up once you apply the `String()` coercion the map already does internally. So the comparison has to go through the tag's idea of key identity somewhere — hard-coding `String()` up in the constructor would be the real leak you want to avoid.

I don't think there's a version that needs zero tag support: either a small opt-in method on the map tags (identity elsewhere), or tracking `overridable` through the tag's own `has`/`get` instead of a raw Set — though that second one gets awkward for `!!set`. Given it's more than a one-liner either way, I'm happy to open an issue and settle on the shape there first, like you suggested.

puzrin:
Yes, create an issue with your vision on "how `{}`-based map should work and why". Prior to make any decision, I'd like to understand the whole situation.

1. Use cases when the user needs non-string scalar keys. Is this really required for all or a rare/specific need? How frequent? And so on...
2. `{}`-based maps are defective by nature (mutate any key kinds to strings). Is this a "bug or feature" :)? Should we continue adding kludges or advise an alternative?
    - I don't know if scalar non-string keys can be considered as really working with `{}` mappings. The user defines a number in the key but gets a string. That's not nice. Is it really needed to enhance features or just say "only string keys are normally supported?
    - We can just advise the `realMap` tag for full-featured key support (it should be ok with any key merge now).
    - Existing parser uses "fast path"; it translates events to JS in a single step. We can suggest a more formal alternative: translate events to AST, then construct JS. Current architecture allows building such chains

IMHO, globally, the question is not "how to make this work", but "how much the user agrees to pay for extra features" (performance, convenience, and so on). I don't think package features should be code-driven instead of user needs. But IMHO, continuing to hack objects may be not the best/optimal approach of available.

Personally, I'd mark Object-based maps as "restricted". But that depends on use cases; I don't know the user needs for YAML.

spokodev:
Opened #778 with the full writeup — the concrete case, the two questions (use cases / bug-or-feature), what `realMapTag` already does with merge, the prior art (#329, #538), and where I'd land. Let's settle the direction there.

puzrin:
Let's close this; changes to merge "as is" are low. This is referred to from an open issue, so any info will not be lost.

PR #777, opened 2026-07-23
fix: quote newline-only strings instead of block scalar

### LINKED ISSUE #534: String "\n" shouldn't be serialized as a block scalar
The string `"\n"` (containing one character, a newline) is serialized as a block scalar:

```yaml
|+


```

This is very awkward to read; it would be much clearer as a double-quoted string.

Also, it breaks my syntax highlighter. While this is undoubtedly a bug with the syntax highlighter, it must be admitted that YAML is very hard to highlight correctly and that js-yaml is not making it any easier in this case. It's also worth noting that the above Markdown code block was wrong the first time I typed it because Markdown ate the significant trailing newline.

This could be fixed in the `chooseScalarStyle` function. I'm not sure what would be a good general rule to avoid this; I'm open to suggestions.

Hashim1999164:
## Summary

When dumping a string that contains only newline characters, the presenter was choosing a literal block scalar (`|+`), which is hard to read for a single `\n`. This change prefers double-quoted style for newline-only strings so `dump('\n')` produces `"\n"`.

## Test plan

- [x] `npm run build`
- [x] `node --test test/core/ast/presenter.test.mjs`

Fixes #534

### DIFF
--- a/src/ast/presenter.ts
+++ b/src/ast/presenter.ts
@@ -433,6 +433,11 @@ function chooseScalarStyle (state: PresenterState, string: string, layout: Retur
   if (blockIndent > 9 && needIndentIndicator(string)) {
     return STYLE_DOUBLE
   }
+  // A string of only newlines dumps poorly as a block scalar (|+ with blank
+  // lines). Prefer a double-quoted escape instead (#534).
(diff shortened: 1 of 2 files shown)


puzrin:
Looks like AI garbage

PR #783, opened 2026-07-29
fix: tag names matching Object.prototype members escape YAMLException

### LINKED ISSUE #782: YAML tag names causing an uncaught TypeError
## Description
YAML tag names are fully user-controlled, via verbatim tags (!<name>) or a %TAG shorthand. js-yaml 5.x resolves them by indexing a plain {} object, so any name that exists on Object.prototype — constructor, toString, __proto__, valueOf, and 8 others — resolves to an inherited member instead of falling through to the intended "unknown tag" error.

The inherited value is truthy, so it is accepted as a tag definition and dereferenced as one. This produces type confusion and throws a raw TypeError. js-yaml documents and otherwise consistently throws YAMLException for malformed input, so this error escapes the error handling of every consumer that follows the documented contract.

## Details
`src/schema.ts:41` builds the tag lookup tables as plain objects, which inherit from `Object.prototype`:

```ts
function createTagDefinitionMap (): TagDefinitionMap {
  return {
    scalar: {},      // <-- inherits from Object.prototype
    sequence: {},
    mapping: {}
  }
}
```

`src/parser/constructor.ts:139` resolves document tag names against them:

```ts
function lookupTag (exact, prefix, tagName) {
  const exactTag = exact[tagName]     // <-- unguarded; tagName comes from the document
  if (exactTag) return exactTag

  for (const tag of prefix) {
    if (tagName.startsWith(tag.tagName)) return tag
  }

  return undefined                    // the intended "unknown tag" path
}
```

`return undefined` is what drives the `unknown ${nodeKind} tag` error in `findExplicitTag`. Inherited members never reach it, because `exact['constructor']` is already truthy.

The confused value is then used as a tag definition:

- **scalar path** → `scalarTag.resolve(...)`
  → `TypeError: scalarTag.resolve is not a function`
- **collection path** → `definition.tag.create(definition.tagName)`
  (`constructor.ts:399`, `:421`)

The collection path is the clearer illustration. For `!<constructor>`, `exact.mapping['constructor']` is the **`Object` constructor**, so `definition.tag.create(definition.tagName)` evaluates to:

```js
Object.create('constructor')
// TypeError: Object prototype may only be an Object or null: constructor
```

## PoS
```js
import { load, YAMLException } from 'js-yaml'

const check = (doc) => {
  try { load(doc); return 'loaded (no error)' } catch (e) {
    return `${e.constructor.name} | instanceof YAMLException: ${e instanceof YAMLException} | ${e.message.split('\n')[0]}`
  }
}

// a tag name that is not registered and not on Object.prototype.
console.log('!<nope>       ', check('!<nope> x'))
//a tag name that IS on Object.prototype.
console.log('!<constructor>', check('!<constructor> x'))
console.log('!<toString>   ', check('!<toString>\n  a: 1\n'))
```

pratik-desgn:
## Summary

Fixes #782. YAML tag names are fully document-controlled (via verbatim tags `!<name>` or a `%TAG` shorthand). The tag lookup tables (`Schema.exact.{scalar,sequence,mapping}`) are built as plain object literals, which inherit from `Object.prototype`. So a tag name that happens to match an inherited member - `constructor`, `toString`, `__proto__`, `valueOf`, and others - resolves to that inherited value in `exact[tagName]` instead of falling through to the intended "unknown tag" error.

The inherited value is truthy, so `lookupTag()` accepts it as a real tag definition and dereferences it as one (`.resolve()` on the scalar path, `.create()` on the collection path), producing a raw `TypeError` that escapes the documented `YAMLException` contract every `load()`/`loadAll()` caller relies on for error handling.

```js
load('!<constructor> x')
// TypeError: scalarTag.resolve is not a function   (should be YAMLException: unknown scalar tag)

load('!<toString>\n  a: 1\n')
// TypeError: definition.tag.create is not a function   (should be YAMLException: unknown mapping tag)
```

## Fix

Build the lookup tables with `Object.create(null)` instead of `{}`, so they have no prototype chain to accidentally shadow the lookup. An unregistered tag name that happens to collide with an `Object.prototype` member now correctly falls through to the same "unknown tag" `YAMLException` as any other unregistered tag name.

## Test plan
- [x] Added 2 regression tests (scalar and mapping paths) asserting `load()` throws `YAMLException` for `!<constructor>` and `!<toString>`. Confirmed both fail with the exact reported `TypeError`s on the old code and pass with the fix.
- [x] Full test suite passes (286/287, 1 pre-existing skip unrelated to this change).
- [x] `tsc --noEmit` and `eslint` both clean.

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -39,10 +39,13 @@ interface TagDefinitionListMap {
 }
 
 function createTagDefinitionMap (): TagDefinitionMap {
+  // Object.create(null) avoids inheriting Object.prototype members (constructor,
+  // toString, __proto__, ...), which would otherwise shadow the "unknown tag" lookup
+  // miss below for a document-controlled tag name that happens to match one of them.
   return {
-    scalar: {},
-    sequence: {},
-    mapping: {}
+    scalar: Object.create(null),
+    sequence: Object.create(null),
+    mapping: Object.create(null)
   }
 }
 
--- a/test/core/units/load-error.test.mjs
+++ b/test/core/units/load-error.test.mjs
@@ -152,6 +152,18 @@ x
   {
     name: 'tab indentation in a later block sequence entry',
     source: ' - foo\n \t- bar\n'
+  },
+  {
+    // An unregistered tag name that collides with an inherited Object.prototype
+    // member (constructor, toString, __proto__, valueOf, ...) must still be
+    // reported as an "unknown tag" YAMLException, not a raw TypeError from
+    // mistakenly dereferencing the inherited value as a tag definition.
+    name: 'unregistered scalar tag matching an Object.prototype member',
+    source: '!<constructor> x'
+  },
+  {
+    name: 'unregistered mapping tag matching an Object.prototype member',
+    source: '!<toString>\n  a: 1\n'
   }
 ]
 


puzrin:
Noise.

https://github.com/nodeca/js-yaml/blob/master/CONTRIBUTING.md

PR #788, opened 2026-08-18
Fix v5 migration guide link in README

aularon:
Updated the link for the v5 migration guide.


brittek:
...

puzrin:
https://github.com/nodeca/js-yaml/commit/04d6fe3298191ee94adf8b1bad87bd2d909092af

I think with the new docs it's simpler to remove that link altogether.

PR #793, opened 2026-08-19
fix: treat a leading UTF-8 BOM as outside the first line

### LINKED ISSUE #791: Sequence parsing in version 5 fails
Since version 5 this rather simple yaml no longer parses corectly:
```yaml
abc: 5
cba:
  - Xyz
  - Zyx
```
```
YAMLException: end of the stream or a document separator is expected (2:1)

 1 | ﻿abc: 5
 2 | cba:
-----^
 3 |   - Xyz
 4 |   - Zyx
```

official-burak:
Fixes #791.

A leading UTF-8 BOM was skipped by advancing `position` only. `lineStart` stayed at 0, so the first content line was treated as indented by one column. A single-key mapping still parsed (`foo: bar`), but a second key at column 0 then failed with `end of the stream or a document separator is expected`.

`consumeLineBreak` already moves `lineStart` with `position` when a line actually starts. The BOM is not content (YAML 1.2.2 [3] `c-byte-order-mark`), so the first line starts after it.

```js
load('\uFEFFabc: 5\ncba:\n  - Xyz\n  - Zyx\n')
// before: YAMLException
// after:  { abc: 5, cba: ['Xyz', 'Zyx'] }
```

`npm test`: 1635 tests, 1622 pass, 0 fail, 13 skipped (the extra passing test is the regression).

### DIFF
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -1449,7 +1449,12 @@ function parseEvents (input: string, options: ParserOptions): Event[] {
   const nullpos = input.indexOf('\0')
   if (nullpos !== -1) YAMLException.throwAt(input, nullpos, 'null byte is not allowed in input', state.filename)
 
-  if (state.input.charCodeAt(state.position) === 0xFEFF) state.position++
+  // UTF-8 BOM is not content. Advance lineStart with position so the first
+  // line is not treated as indented by one (YAML 1.2.2 §5.2 / [3] c-byte-order-mark).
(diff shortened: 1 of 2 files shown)


puzrin:
https://github.com/nodeca/js-yaml/blob/master/CONTRIBUTING.md

Don't PR without agreement on the scope, please. That's fixed in a better way.

official-burak:
Thank you.

PR #794, opened 2026-08-19
fix: honor sortKeys when deciding if a document is open-ended

### LINKED ISSUE #789: `isOpenEnded()` ignores mapping order used by `sortKeys`
When `sortKeys` changes which mapping value is rendered last, `isOpenEnded()` still uses the original item order. Its open-ended decision can therefore disagree with the document that `present()` actually renders.

The renderer sorts with `sortMappingItems()`, while `isOpenEnded()` reads `node.items[node.items.length - 1]` directly. It should use the same order as the renderer when deciding whether a document is open-ended.

```mjs
import { CORE_SCHEMA, eventsToAst, jsToAst, parseEvents, present } from '../../repo/src/index.ts'

const source = 'b: |+\n  x\n\na: x\n'
const documents = eventsToAst(parseEvents(source, {}), { source, schema: CORE_SCHEMA })
documents.push(...jsToAst('next', CORE_SCHEMA))

const output = present(documents, { schema: CORE_SCHEMA, sortKeys: true })
console.log(output)
```

output

```
a: x
b: |+
  x

--- next
```

With the rendered order, the final mapping value is the keep-chomped scalar b, so isOpenEnded() should return true. Under the presenter's existing open-ended handling, this would emit ... before the next document rather than ---.

```
a: x
b: |+
  x

...
next
```

official-burak:
Fixes #789.

`writeBlockMapping` renders keys in `sortKeys` order, but `isOpenEnded()` still looked at `node.items[length - 1]`. If sorting moved a keep-chomped (`|+`) scalar to the end, the presenter did not emit `...` before the next document.

```js
const source = 'b: |+\n  x\n\na: x\n'
const documents = eventsToAst(parseEvents(source, {}), { source, schema: CORE_SCHEMA })
documents.push(...jsToAst('next', CORE_SCHEMA))
present(documents, { schema: CORE_SCHEMA, sortKeys: true })
// before: --- next
// after:  ...\nnext
```

`npm test`: 1635 tests, 1622 pass, 0 fail, 13 skipped.

### DIFF
--- a/src/ast/presenter.ts
+++ b/src/ast/presenter.ts
@@ -1005,15 +1005,17 @@ function rootStartsOwnLine (node: Node) {
 // A document whose serialization ends with a keep-chomped (`+`) block scalar is
 // open-ended: the trailing blank line(s) would otherwise be ambiguous, so it
 // needs a `...` terminator. Mirrors the keep test in `blockHeader`.
-function isOpenEnded (node: Node) {
+function isOpenEnded (node: Node, state: PresenterState) {
   // Descend to the last leaf, always taking the last item of a block collection
   // (a flow collection renders on one line, so it ends the document itself).
+  // Mapping order must match writeBlockMapping: sortKeys can change which
+  // value is actually rendered last (#789).
   let leaf = node
   while ((leaf.kind === 'sequence' || leaf.kind === 'mapping') &&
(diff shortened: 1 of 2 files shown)


official-burak:
Thank you. Understood, and noted regarding CONTRIBUTING. Closing this PR.

PR #797, opened 2026-08-26
fix: count empty merge sources against maxTotalMergeKeys

spokodev:
The `maxTotalMergeKeys` guard added for the quadratic merge-key DoS (GHSA-g796-fgmg-93mv, fixed in 5.2.0) charges the budget **once per folded key**. A merge source that folds **no** keys — an empty mapping — costs real work to process but never increments the counter, so the guard can be bypassed:

```yaml
arr: &arr [{}, {}, {}, ...]   # N empty mappings
targets:
  - <<: *arr                  # repeated K times
```

`mergeSource` iterates every element of the `<<:` sequence, and aliasing one N-element array of empty maps into K targets does `O(K*N)` work — none of it counted.

## Impact

With merge enabled (`YAML11_SCHEMA`, or `CORE_SCHEMA.withTags(mergeTag)` — the standard YAML-1.1 / v4-compat configuration), the load time scales quadratically:

| payload | time |
|---|---|
| N=K=800 (~13 KB) | ~20 ms |
| N=K=3200 (~50 KB) | ~180 ms |
| N=K=20000 (~500 KB) | **~13 s** |

A ~1 MB document projects to minutes of CPU on a single core. This restores the same availability DoS the `maxTotalMergeKeys` guard was shipped to prevent. A **real** key-folding merge of comparable size is correctly rejected today (`merge keys exceeded maxTotalMergeKeys`), so the guard works for the intended shape but is defeated by zero-key sources.

**Attacker → control:** attacker supplies an untrusted YAML document to an app that loads with merge enabled; the defeated control is `maxTotalMergeKeys` (default 10000).

## Fix

Track whether a source folded any keys; if it folded none, charge one unit for it. A sequence of empty mappings is then bounded by `maxTotalMergeKeys` like any other merge.

## Tests

A case in `test/core/pathological.test.mjs` (next to the existing merge-chain test) asserts the empty-source payload is rejected; it hangs for ~13 s and loads on the current code, and throws promptly with the fix. Normal and single empty-map merges still load correctly. Full suite green.

### DIFF
--- a/src/parser/constructor.ts
+++ b/src/parser/constructor.ts
@@ -237,13 +237,20 @@ function isMappingTag (tag: AnyTag): tag is MappingTagDefinition<any, any> {
   return tag.nodeKind === 'mapping'
 }
 
+function chargeMergeKey (state: ConstructorState) {
+  if (state.maxTotalMergeKeys !== -1 && ++state.totalMergeKeys > state.maxTotalMergeKeys) {
+    throwError(state, `merge keys exceeded maxTotalMergeKeys (${state.maxTotalMergeKeys})`)
+  }
+}
+
 // Fold the keys of one mapping source into the target frame, honoring merge
 // precedence: an already-present key (explicit or from an earlier source) wins.
 function mergeKeys (state: ConstructorState, frame: MappingFrame, source: unknown, sourceTag: MappingTagDefinition<any, any>) {
+  let counted = false
+
   for (const sourceKey of sourceTag.keys(source)) {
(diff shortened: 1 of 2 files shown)


puzrin:
Note: next time, things will be simpler if you report such a problem via the standard way instead of a PR https://github.com/nodeca/js-yaml?tab=security-ov-file.

The cost of writing code is much lower than that of PR processing. Especially when a PR needs to be reviewed urgently and there's no option to request improvements.

puzrin:
See https://github.com/nodeca/js-yaml/commit/dfd3a2933c69959fc16fce1a775f36c9a5c89063

